PR #1, opened 2023-02-21
Brad Dunlap Mid Mod

brad-dunlap:
I made it through all user stories and the extension. This one was a bit tougher and I had some trouble with model methods, but I think I got it in the end.


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,12 @@
+class GuestsController < ApplicationController
+	def show
+		@guest = Guest.find(params[:id])
+	end
+
+	def update
+		@guest = Guest.find(params[:id])
+    @room = Room.find(params[:room_id])
+    @guest.rooms << @room
```
cjsim89, line 9: The functionality of what you're doing here is correct, we're adding an existing Room to a Guest. However, this action of _creating_ something is happening within an `update` method. I think a more RESTful approach would be to have this in its own `create` method, inside of a `RoomGuestsController`. That way, that specific controller is responsible for the CRUD actions for relationships between Rooms and Guests.

cjsim89: Basically lifting this method and calling it `create` within a `RoomGuestsController` would be perfect.

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+	def show
+		@hotel = Hotel.find(params[:id])
+		@unique = @hotel.unique_guests
```
cjsim89, line 4: NIce job making a method here. You can also call this method directly from the View, you don't necessarily have to make another instance variable for it.

app/models/hotel.rb
```diff
@@ -1,3 +1,8 @@
 class Hotel < ApplicationRecord
    has_many :rooms
+	 has_many :guests, through: :rooms
```
cjsim89, line 3: Good work making this association! 🙌

app/models/hotel.rb
```diff
@@ -1,3 +1,8 @@
...
+
+	 def unique_guests
+		guests.distinct
```
cjsim89, line 6: Based on that association you can make simple AR like this. Nice work

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<p> Rooms Stayed In:</p>
+<ul>
+<% @guest.rooms.each do |room| %>
+<li> <%= room.suite %> ($<%= room.rate %> at the <%= room.hotel.name %>) </li>
+<% end %>
+</ul>
+</div>
+
+<div id="add_rooms">
+	<%= form_with url: "/guests/#{@guest.id}", method: :patch, local: true do |form| %>
```
cjsim89, line 13: I think the first part of this path is ok, but i disagree that we're doing an `update` functionality here, we're really creating a new relationship between an existing Room and an existing Guest. 

We'll need the `guest_id` in the path or in the request body of the form, but the URL could also be: 
```
/guests/:guest_id/rooms
```
and then we send in the `room_id` in the form that way. Also, the method should be `:post` since we're creating a relationship, not updating either Guest or Room itself.

cjsim89: The path could also potentially be
```
/room_guests
```
with a `:post` method. If you did that, you'd have to include a `form.hidden_field` whose value is the `guest_id`.

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1> <%= @hotel.name %></h1>
+<div id="hotel">
```
cjsim89, line 2: Cool to use a div here; you can also assign the `<ul>` element an ID, to save some typing.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<ul>
+<% @rooms.each do |room| %>
+<div id="room-<%= room.id %>"> 
+<li>Suite: <%= room.suite %>
+<br>
+Nightly Rate: $<%= room.rate %>
+<br>
+Hotel: <%= room.hotel.name %>
+<br>
+Total Guests: <%= room.guests.count %>
```
cjsim89, line 12: This is the right functionality, but this`.count` here is technically some AR/SQL since it modifies the query. It's better practice to have this in a model method.

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+	
+  scenario 'see a unique list of all guests that have stayed at this hotel.' do
+		visit "/hotels/#{@hotel.id}"
+		save_and_open_page
+		within "div#hotel" do
+
+    	expect(page).to have_content("Brad Dunlap")
+    	expect(page).to have_content("Adrian Fierro")
+    	expect(page).to have_content("Aaron McKendry")
+		end
```
cjsim89, line 30: Just a note about readability - maybe it's your linter or an extension in your VS Code, but some of the indentations are off in this PR and it makes it somewhat difficult to read in places. This will be important to notice and correct more in group/paired settings, just keep an eye out that your `do/end`s are all in-line with each other, and the containing lines are indented to the right of the do/end.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+		RoomGuest.create!(guest: @guest3, room: @pent)
+		visit '/rooms'
+	end
+	
+  scenario 'see a list of all rooms' do
+		within 'div#list_of_rooms' do
+			expect(page).to have_content("Presidential")
+			expect(page).to have_content("Penthouse")
+			expect(page).to have_content("Suite")
+			expect(page).to have_content("Honeymoon Suite")
```
cjsim89, line 27: All of your tests have great uses for `within` blocks. One way to make them stronger would be to use at least one `to_not`/ negative case in each `within`. That way you can be extra sure that only what is supposed to be in this part of the page is there.

cjsim89 (commented):
Overall your testing is pretty strong, there were 2 MVC infractions that I saw - one where using `:patch` to create a relationship (should be `:post` instead), and using `.count` in your view... That one's pretty tricky but it is technically an aggregator/calcluation method that should be done in a model method, and then tested. Keep an eye out for those and you will be unstoppable 💪

PR #2, opened 2023-02-21
Hady Matar

hadyematar23:
I thought this went fine!

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,32 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+    guest_room = GuestRoom.new(guest_room_params)
+
+    if guest_room.save 
+      redirect_to "/guests/#{guest_room.guest_id}"
+    end
+  end
+
+  private 
+
+  def guest_room_params
+    params.permit(:guest_id, :room_id)
+  end
+end 
+
+
+# def create 
+#   application = Application.new(application_params)
+
+#   if application.save
+#     redirect_to "/applications/#{application.id}" 
+#   else
+#     redirect_to "/applications/new"
+#     @errors = application.errors.messages
+#     flash[:alert] = @errors.map do |error|
+#       "#{error.first.capitalize} #{error.last}.".gsub(/[\["\]]/, "")
+#     end.join.gsub(/\./, ". ")
+#   end 
+# end
+
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms 
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,4 +1,8 @@
 class Guest < ApplicationRecord
   has_many :guest_rooms 
   has_many :rooms, through: :guest_rooms
+
+  def self.unique_list_of_guests(id)
+    joins(rooms: :hotel).where(hotels: {id: id}).distinct
+  end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -5,3 +5,10 @@
     <p><%=("#{@guest.name} stayed in the #{room.suite} at #{room.hotel.name} for #{room.rate} per night.")%></p>
   <%end%>
 </div>
+
+<div id="add_room_to_guest">
+  <%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |form| %>
+    <%= form.number_field "room_id" %>
+    <%= form.submit "Add Room to Guest" %>
+  <%end %>
+</div>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,7 @@
+<h1><%="Guest Name: #{@guest.name}"%></h1>
+
+<div id="rooms_stayed_in">
+  <%@guest.rooms.each do |room|%>
+    <p><%=("#{@guest.name} stayed in the #{room.suite} at #{room.hotel.name} for #{room.rate} per night.")%></p>
+  <%end%>
+</div>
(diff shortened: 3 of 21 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,32 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+    guest_room = GuestRoom.new(guest_room_params)
+
+    if guest_room.save
```
cjsim89, line 6: Nice job with the `.save` and using strong params; this leaves you set up for sad paths eventually. 👍

app/models/guest.rb
```diff
@@ -0,0 +1,8 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms 
+  has_many :rooms, through: :guest_rooms
+
+  def self.unique_list_of_guests(id)
```
cjsim89, line 5: This works.... but if you're passing in a Hotel ID, I would call your parameter name this as well, to distinguish it from a Guest ID (just a little confusing to read).

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1><%="Guest Name: #{@guest.name}"%></h1>
+
+<div id="rooms_stayed_in">
+  <%@guest.rooms.each do |room|%>
+    <p><%=("#{@guest.name} stayed in the #{room.suite} at #{room.hotel.name} for #{room.rate} per night.")%></p>
+  <%end%>
+</div>
+
+<div id="add_room_to_guest">
+  <%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |form| %>
```
cjsim89, line 10: This path is ok, but I think it could be more RESTful if the `guest_id` were in a hidden field in the form. This would mean the end of the path matches with the `:post` method.

config/routes.rb
```diff
@@ -1,3 +1,12 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  get "/guests/:id", to: "guests#show" 
+
+  post "/guest_rooms/:guest_id", to: "guest_rooms#create"
```
cjsim89, line 6: I'm ok with this route, but overall if a route is creating something, it won't have an ID in the path; if we're creating a relationship from Guest to Room, then both could be contained within the form data sent along in the request body.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        expect(page).to have_button("Add Room to Guest")
+
+      end
+
+      it "you fill in the field with the id of an existing room, click submit and are redirected back to the guest show page where you see the newly added room now listed under the guest room" do 
+
+        visit "/guests/#{@guest_hady.id}"
+
+        within("div#rooms_stayed_in") do
+          expect(page).to_not have_content(@room_san_jose.suite)
```
cjsim89, line 50: This is a really well-written test! I like how you test the `to_not` case first, which gives more weight to the one later on where we do expect to see this content. Keep this up

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  GuestRoom.create!(guest_id: @guest_hady.id, room_id: @room_jocotopec.id)
+  GuestRoom.create!(guest_id: @guest_hady.id, room_id: @room_da_flow.id)
+
+  GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_jocotopec.id)
+  GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_da_flow.id)
+
+  GuestRoom.create!(guest_id: @guest_sebastian.id, room_id: @room_da_flow.id)
+  GuestRoom.create!(guest_id: @guest_sebastian.id, room_id: @room_san_jose.id)
+
+  GuestRoom.create!(guest_id: @guest_axel.id, room_id: @room_da_flow.id)
```
cjsim89, line 28: Really good test setup with objects, and nice job keeping it efficient with the `before :each`

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe "as a visitor" do
+    describe "when visit the hotel show page" do 
+      it "displays a unique list of all the guests at the hotel" do
+
+        visit "/hotels/#{@hotel_1.id}"
+
+        within("div#unique_list_of_guests") do 
+          expect(page).to have_content(@guest_hady.name)
+          expect(page).to have_content(@guest_malena.name)
+          expect(page).to have_content(@guest_sebastian.name)
```
cjsim89, line 42: Something else useful in feature tests, is if you want to be sure this list is unique, you could do 
```
expect(page).to have_content(@guest_hady.name, count: 1)
```
so that only one instance of that string/content is found in the page.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_jocotopec.id)
+      GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_da_flow.id)
+    
+      GuestRoom.create!(guest_id: @guest_sebastian.id, room_id: @room_da_flow.id)
+    
+      GuestRoom.create!(guest_id: @guest_axel.id, room_id: @room_da_flow.id)
+    end 
+
+    it "unique_list_of_guests(id)" do 
+      expect(Guest.unique_list_of_guests(@hotel_1.id)).to eq([@guest_malena, @guest_hady])
```
cjsim89, line 37: Great test - I think if you also have `@hotel_2` in this context, it's good to test that the value of the method is different than `@hotel_1`.

spec/models/room_spec.rb
```diff
@@ -3,5 +3,37 @@
...
+    
+      GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_jocotopec.id)
+      GuestRoom.create!(guest_id: @guest_malena.id, room_id: @room_da_flow.id)
+    
+      GuestRoom.create!(guest_id: @guest_sebastian.id, room_id: @room_da_flow.id)
+      GuestRoom.create!(guest_id: @guest_sebastian.id, room_id: @room_san_jose.id)
+    
+    end 
+    it "#count_guests" do 
+      expect(@room_da_flow.count_guests).to eq(3)
```
cjsim89, line 36: Similar to the other model test, I think it's good since you have multiple room instances to work with, to write a few more expect lines for other rooms. This helps to ensure the model returns different values in different scenarios, makes your app overall stronger.

PR #4, opened 2023-02-21
Meredith Trevino

MATrevino:
Reflection: I feel ok about this IC, however I'm not fully confident in my WHY behind some decisions/coding. I need to go back and review why I need to follow certain steps. I want to feel more confidence in my coding in order to help others and help myself feel better about explaining my own work.

Completed: user stories 1-3 and extension

A random question I had- When I started on user story 3, my capybara/visit request was throwing an error so I added "  config.include Capybara::DSL " inside my rails_helper as advised by stackoverflow. It worked and fixed whatever issue there was, however what was the issue? the visit method was working on my previous stories.


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```
abdulredd, line 7: 👍

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @hotel_guests = @hotel.list_of_guests
+  end
+end
```
abdulredd, line 6: 👍

app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,9 @@
+class RoomGuestsController < ApplicationController
+
+  def create
+    room_guest = RoomGuest.new(room_id: params[:room_id], guest_id: params[:id])
+    
+    room_guest.save
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 9: 👍

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,6 @@
+class Guest < ApplicationRecord
+  has_many :room_guests
+  has_many :rooms, through: :room_guests
+
+
+end
```
abdulredd, line 6: 👍

app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
    has_many :rooms
+
+  def list_of_guests
+    Guest.joins(:room_guests).distinct.pluck(:name)
+  end
 end
```
abdulredd, line 7: I really the `list_of_guests` method here. What if I told you there was an easier way to do this via an association?

```ruby
  has_many :rooms

  def list_of_guests
    guests.distinct
  end
end
```

Rails magic FTW. Play around with this and see what SQL gets printed in the terminal. You could the name or whatever you wanted to do. Reach out if you want to discuss this further.

app/models/room.rb
```diff
@@ -1,3 +1,6 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+
+   has_many :room_guests
+   has_many :guests, through: :room_guests
 end
```
abdulredd, line 6: 👍

app/models/room_guest.rb
```diff
@@ -0,0 +1,4 @@
+class RoomGuest < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <h4>Name of hotel: <%=room.hotel.name%></h4>
+  <br>
+  <%end%>
+
+<%= form_with url: "/guests/#{@guest.id}/room_guests", method: :post, local: :true do |form|%>
+<%= form.label "Room ID:"%>
+<%= form.number_field :room_id%>
+<%= form.submit "Submit"%>
+
+<%end%>
```
abdulredd, line 16: I will touch on this in the `routes.rb` file, but the URL for the form isn't correct according to REST.

```ruby
```

```ruby
<%= f.label :room_id,  "Room ID: " %>
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,5 @@
+<h2><%=@hotel.name%></h2>
+<p>List of guests who have stayed at <%=@hotel.name%> </p>
+<%@hotel_guests.each do |guest|%>
+<p><%=guest%></p>
+<%end%>
```
abdulredd, line 5: 👍

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<%@rooms.each do |room|%>
+<h2>Suite Name: <%=room.suite%></h2>
+<p>Nightly Rate: $<%=room.rate%></p>
+<p>Hotel: <%=room.hotel.name%></p>
+<p>Number of guests who have stayed: <%=room.guests.count%></p>
+<%end%>
```
abdulredd, line 6: If we think about line `5` where you are doing `room.guests.count` and think about what it's doing. On that line, it's taking `room.guests` and then calling the ActiveRecord aggregate method `count` on it. This is performing a database call, and because this is hitting the database, this code needs to be in the `Room` model.

I just want to compare that to line `4`, where you are doing `room.hotel.name`. On this line, you aren't calling the database because when we called on all the `@rooms` in the controller, those rooms, came with the hotel data that is linked to each room, and each of those hotels came with its attributes, including the `name`. No additional database call is being made there.

config/routes.rb
```diff
@@ -1,3 +1,11 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+get '/guests/:id', to: "guests#show"
+
+post '/guests/:id/room_guests', to: "room_guests#create"
+
+get '/rooms', to: "rooms#index"
+
+get '/hotels/:id', to: "hotels#show"
+
```
abdulredd, line 10: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221160459_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221161657_create_room_guests.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRoomGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :room_guests do |t|
+      t.references :room, foreign_key: :true
+      t.references :guest, foreign_key: :true
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 9: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        click_on "Submit"
+
+        expect(current_path).to eq("/guests/#{guest1.id}")
+        expect(page).to have_content("Suite: Balcony Standard")
+        expect(page).to have_content("Rate: $195")
+        expect(page).to have_content("Name of hotel: Snowed Inn")
+      end
+    end
+  end
+end
```
abdulredd, line 56: 👍 

```ruby
```

Beyond that, the tests are set up perfectly, and I like how you used full sentence strings in your assertions to give them a little more weight and less ambiguity.

The only recommendation other recommendation I would make is to start working in `within` blocks into your tests.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      expect(page).to have_content('Hotel: Echo Mountain Inn')
+      expect(page).to have_content("Hotel: Snowed Inn")
+      expect(page).to have_content("Hotel: Loveland Hotel")
+
+      expect(page).to have_content("Number of guests who have stayed: 1")
+      expect(page).to have_content("Number of guests who have stayed: 2")
+    end
+  end
+end
```
abdulredd, line 44: Another great test!

spec/hotels/show_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      visit "/hotels/#{hotel1.id}"
+      
+      expect(page).to_not have_content("Loveland Hotel")
+      expect(page).to have_content("List of guests who have stayed at #{hotel1.name}")
+      expect(page).to have_content("Charlize Theron")
+      expect(page).to have_content("Tim Allen")
+    save_and_open_page
+    end
+  end
+end
```
abdulredd, line 30: The way the test is set up doesn't fully cover what is expected in the user story. Right now you are checking that `"Charlize Theron"` and `"Tim Allen"` show up in the browser but not checking if they show up just once. This is what you want:

```ruby
expect(page).to have_content("Charlize Theron").once
expect(page).to have_content("Tim Allen").once
```
or
```ruby
expect(page).to have_content("Charlize Theron", count: 1)
expect(page).to have_content("Tim Allen", count: 1)
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it {should have_many :room_guests}
+    it {should have_many :rooms}
+  
+  end
+end
```
abdulredd, line 9: 👍

spec/models/hotel_spec.rb
```diff
@@ -1,7 +1,26 @@
...
   describe 'relationships' do
     it {should have_many :rooms}
   end
+
+  describe 'instance methods' do
+    it '#list_of_guests' do
+      expect(hotel1.list_of_guests).to eq(["Tim Allen", "Charlize Theron"])
+    end
+  end
 end
```
abdulredd, line 26: 👍

spec/models/room_guest_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe RoomGuest, type: :model do
+  describe 'relationships' do
+    it {should belong_to :room}
+    it {should belong_to :guest}
+   
+  end
+end
```
abdulredd, line 9: 👍

spec/models/room_spec.rb
```diff
@@ -3,5 +3,9 @@
 RSpec.describe Room, type: :model do
   describe 'relationships' do
     it {should belong_to :hotel}
+    it {should have_many :room_guests}
+    it {should have_many :guests}
+
+
   end
 end
```
abdulredd, line 11: 👍

spec/rails_helper.rb
```diff
@@ -43,6 +43,7 @@
 
   # You can uncomment this line to turn off ActiveRecord support entirely.
   # config.use_active_record = false
+  config.include Capybara::DSL
 
   # RSpec Rails can automatically mix in different behaviours to your tests
   # based on their file location, for example enabling you to call `get` and
```
abdulredd, line 49: To answer your reflection question about the reason why `visit` wasn't working in your `hotels/show_spec.rb` file. 

If you remember, we said that if you wanted, you could add this to your feature tests: `type: :feature`, and if you didn't Rails would determine what type of test it was by looking at the folder where the test was located. Well in your `hotels/show_spec.rb` file, you didn't specify the `type`, which is ok, but you also didn't put `hotels/` in the `features/` folder so Rails was unable to determine what type of test this was. `visit` is only available in feature tests.

Moving `hotels/` into `features/` was all you needed. You don't need the `config.include Capybara::DSL` line.

PR #5, opened 2023-02-21
Conner Van Loan Mid Mod IC

C-V-L:
This PR completes all User Stories and the Extension.
I felt that this IC went relatively smooth. I made an error on my initial setup where I used `rails g model createGuests` and this resulted in a table named "create_guests". I attempted to resolve the naming a few different ways but ultimately had to delete the repo from my local machine and re-clone down the git repo. I need to practice altering tables via command line/migrations. Aside from that error, I was able to move quickly through the stories and extension.

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    GuestRoom.create!(guest: Guest.find(params[:id]), room: Room.find(params[:room_id]))
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,5 +1,6 @@
 class GuestsController < ApplicationController
   def show
     @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
   end
 end
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @guests = @hotel.guests.distinct
+  end
+end
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -1,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms, dependent: :destroy
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,4 +1,6 @@
 class Guest < ApplicationRecord
   has_many :guest_rooms, dependent: :destroy
   has_many :rooms, through: :guest_rooms
+
+
 end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
-   has_many :rooms
+  has_many :rooms
+  has_many :guest_rooms, through: :rooms
+  has_many :guests, through: :guest_rooms
+
+
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -8,6 +8,6 @@ class Room < ApplicationRecord
   end
 
   def guest_count
-    guests.count
+    guests.distinct.count
   end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,8 @@ class Room < ApplicationRecord
    belongs_to :hotel
    has_many :guest_rooms, dependent: :destroy
    has_many :guests, through: :guest_rooms
+
+   def hotel_name
+      Hotel.find(hotel_id).name
+    end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms, dependent: :destroy
+   has_many :guests, through: :guest_rooms
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -5,5 +5,9 @@ class Room < ApplicationRecord
 
    def hotel_name
       Hotel.find(hotel_id).name
-    end
+  end
+
+  def guest_count
+    guests.count
+  end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -6,3 +6,10 @@
   <%= "Rate: #{room.rate}" %> <br>
   <%= "Hotel: #{room.hotel_name}" %> <br><br>
 <% end %>
+
+<h2> Add a Room for this Guest</h2>
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label :room_id, "Room ID" %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Add Room" %>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,1 @@
+<h1> <%= "Guest: #{@guest.name}" %> </h1>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,1 +1,8 @@
 <h1> <%= "Guest: #{@guest.name}" %> </h1>
+
+<h2> Rooms Stayed In </h2>
+<h3><% @rooms.each do |room| %></h3>
+  <%= room.suite %> <br>
+  <%= "Rate: #{room.rate}" %> <br>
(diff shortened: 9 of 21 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestRoomsController < ApplicationController
+  def create
+    GuestRoom.create!(guest: Guest.find(params[:id]), room: Room.find(params[:room_id]))
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 6: I don't have an issue with this, just looks a little "off". I would much rather you take one of the 2 following approaches:

```ruby
guest = Guest.find(params[:id])
room = Room.find(params[:room_id])
GuestRoom.create!(guest: guest, room: room)
```
or
```ruby
GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
```

Although this does the exact same thing yours does, I think it's a little better visually. (My opinion)

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @guests = @hotel.guests.distinct
+  end
+end
```
abdulredd, line 6: Using `distinct` in your controller is performing database logic, and by having it in the controller, we don't get the opportunity to test this method.

`guests.distinct` should really be in the model and then the call can be called from the controller.

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,6 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms, dependent: :destroy
+  has_many :rooms, through: :guest_rooms
+
+
+end
```
abdulredd, line 6: 👍

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+end
```

app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
-   has_many :rooms
+  has_many :rooms
+  has_many :guest_rooms, through: :rooms
+  has_many :guests, through: :guest_rooms
+
+
 end
```
abdulredd, line 7: 👍

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
...
+   has_many :guests, through: :guest_rooms
+
+   def hotel_name
+      Hotel.find(hotel_id).name
+  end
+
+  def guest_count
+    guests.distinct.count
+  end
 end
```
abdulredd, line 13: In your `hotel_name` method, I wonder why you needed to find the hotel. We already have that association set up, `belongs_to :hotel`. So our logic could have just been, `hotel.name`.

Beyond that, you don't need to necessarily create a new method to get the name of a hotel because this logic is not making a call to the database, it's just retrieving the name attribute on the hotel object. You could have just called `room.hotel.name` from the view with no issue.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= "Rate: #{room.rate}" %> <br>
+  <%= "Hotel: #{room.hotel_name}" %> <br><br>
+<% end %>
+
+<h2> Add a Room for this Guest</h2>
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label :room_id, "Room ID" %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Add Room" %>
+<% end %>
```
abdulredd, line 15: We mustn't forget about the HTML. Always put rendered ERB and plaintext in an HTML tag.

```ruby
<p><%= room.suite %></p>

<p><%= "Rate: #{room.rate}" %></p>

<p><%= "Hotel: #{room.hotel_name}" %></p>

```

---

I will touch on this more in the `routes.rb` file, but the URL for the form isn't correct according to REST.

```ruby
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1> <%= @hotel.name %> </h1>
+
+<h2> Guests: </h2>
+<% @guests.each do |guest| %>
+  <%= guest.name %> <br>
+<% end %>
```
abdulredd, line 6: Missing HTML

```ruby
<p><%= guest.name %></p>
```

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Rooms</h1>
+
+<% @rooms.each do |room| %>
+  <h2><%= "Room: #{room.suite}" %> <br></h2>
+  <%= "Rate: #{room.rate}" %> <br>
+  <%= "Hotel: #{room.hotel_name}" %> <br>
+  <%= "Number of Guests Stayed: #{room.guest_count}" %>
+<% end %>
```
abdulredd, line 8: Missing HTML

```ruby
<p><%= "Rate: #{room.rate}" %></p>
<p><%= "Hotel: #{room.hotel_name}" %></p>
<p><%= "Number of Guests Stayed: #{room.guest_count}" %></p>
```

config/routes.rb
```diff
@@ -1,3 +1,11 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get "guests/:id", to: 'guests#show'
+
+  post "guests/:id/guest_rooms", to: 'guest_rooms#create'
+
+  get "rooms", to: 'rooms#index'
+
+  get "hotels/:id", to: 'hotels#show'
+
```
abdulredd, line 10: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221161805_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221161941_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.belongs_to :guest, foreign_key: true
+      t.belongs_to :room, foreign_key: true
+      
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        fill_in "Room ID", with: @castleton.id
+        click_button "Add Room"
+        expect(current_path).to eq "/guests/#{@guest_1.id}"
+        expect(page).to have_content "Castleton"
+        expect(page).to have_content "Rate: 175 Hotel: Red Cliffs Lodge"
+      end
+    end
+  end
+end
```
abdulredd, line 46: 👍 

The only recommendation I would make here is to add some "not have" assertions.

```ruby
expect(page).to_not have_content "Castleton"
expect(page).to_not have_content "Rate: 175 Hotel: Red Cliffs Lodge"
```

Beyond that, working in some `within` blocks would give your tests more weight as well. But with the way your tests are currently constructed with full string assertions, I like them the way they currently are also.

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      visit "/hotels/#{@red_cliffs.id}"
+
+      expect(page).to have_content "Guests:"
+      expect(page).to have_content "Nik"
+      expect(page).to have_content "Matt"
+      expect(page).to_not have_content "Jimmy"
+    end
+  end
+end
```
abdulredd, line 30: The way the test is set up it doesn't fully cover what is expected in the user story. Right now you are checking that `"Nik"` or `"Matt"` show up in the browser but not checking if they show up just once. This is what you want:

```ruby
expect(page).to have_content("Nik").once
expect(page).to have_content("Matt").once
```
or
```ruby
expect(page).to have_content "Nik", count: 1
expect(page).to have_content "Matt", count: 1
```

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it 'lists the number of guests that have stayed in each room' do 
+      visit "/rooms"
+
+      expect(page).to have_content "Number of Guests Stayed: 2"
+      expect(page).to have_content "Number of Guests Stayed: 1"
+      expect(page).to have_content "Number of Guests Stayed: 0"
+
+    end
+  end
+end
```
abdulredd, line 42: 👍

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  # pending "add some examples to (or delete) #{__FILE__}"
+end
```
abdulredd, line 5: You are missing your relationship/association tests.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  # pending "add some examples to (or delete) #{__FILE__}"
+end
```
abdulredd, line 5: You are missing your relationship/association tests.

spec/models/room_spec.rb
```diff
@@ -4,4 +4,33 @@
...
+      expect(@cabin.guest_count).to eq 1
+
+      GuestRoom.create!(guest: @guest_1, room: @cabin)
+      expect(@cabin.guest_count).to eq 2
+
+      GuestRoom.create!(guest: @guest_1, room: @cabin)
+      expect(@cabin.guest_count).to eq 2
+    end
+  end
 end
```
abdulredd, line 36: You are missing the rest of your relationship/association tests.

PR #6, opened 2023-02-21
Kara Jones-Hofmann

KaraJoHo:
Got through the extension. Not too bad. I have to take my sweet time setting up the migrations though. Other than that, it was good

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,3 +1,13 @@
 class GuestRoomsController < ApplicationController 
-  
+  def create 
+    new_guest_room = GuestRoom.create!(guest_room_params)
+
+    redirect_to "/guests/#{params[:guest_id]}"
+  end
+
+  private 
+
+  def guest_room_params 
+    params.permit(:guest_id, :room_id)
+  end
 end
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,3 @@
+class GuestRoomsController < ApplicationController 
+  
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord 
+  has_many :guest_rooms 
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,4 +1,6 @@
 class Guest < ApplicationRecord 
   has_many :guest_rooms 
   has_many :rooms, through: :guest_rooms
+  validates_presence_of :name
+  validates_numericality_of :nights
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,9 +1,19 @@
 <h1> <%= @guest.name %> </h1>
 
-<% @guest.rooms.each do |room| %>
-  <div id = <%="#{room.id}"%>>
-    <h3>Suite: <%= room.suite %></h3>
-    <p>Nightly Rate: <%= room.rate %>$</p><br>
-    <p>Hotel: <%= room.hotel.name %></p><br>
-  </div>
-<% end %>
+<div class = "guest_rooms">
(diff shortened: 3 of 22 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,13 @@
+class GuestRoomsController < ApplicationController 
+  def create 
+    new_guest_room = GuestRoom.create!(guest_room_params)
```
cjsim89, line 3: `new_guest_room` is fine to create as a local variable, but since you're not using it anywhere, you shouldn't have to assign it at all. Just doing `GuestRoom.create!...` is fine for these relationships.

app/models/guest.rb
```diff
@@ -0,0 +1,6 @@
+class Guest < ApplicationRecord 
+  has_many :guest_rooms 
+  has_many :rooms, through: :guest_rooms
+  validates_presence_of :name
+  validates_numericality_of :nights
```
cjsim89, line 5: Nice job validating your data! Just so you are aware, we won't ask you to do this for the final or other ICs, just projects.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <div id = <%="#{room.id}"%>>
+      <h3>Suite: <%= room.suite %></h3>
+      <p>Nightly Rate: <%= room.rate %>$</p><br>
+      <p>Hotel: <%= room.hotel.name %></p><br>
+    </div>
+  <% end %>
+</div>
+
+<div class = "add_room_form">
+  <%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |form| %>
```
cjsim89, line 14: Excellent path/verb combo here, it's exactly what I would have done 🙌

cjsim89: An alternative that might be easier to read would be
```
url: "/guests/:guest_id/rooms"
```
and then send in the room Id as you have done in the form.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'rails_helper'
+
+RSpec.describe "Guest Show Page" do 
+  before(:each) do
```
cjsim89, line 4: Nice job using a `before :each`, and you have really good test data to work from 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        expect(page).to_not have_content(@hotel_2.name)
+        expect(page).to_not have_content(@room_4.rate)
+      end
+
+      within("##{@room_2.id}") do 
+        expect(page).to have_content(@room_2.suite)
+        expect(page).to have_content(@room_2.rate)
+        expect(page).to have_content(@hotel_1.name)
+        expect(page).to_not have_content(@hotel_2.name)
+        expect(page).to_not have_content(@room_4.rate)
```
cjsim89, line 48: Excellent `within` blocks and `to_not` examples inside 👍  Keep this up

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    @guest_rooms_4 = GuestRoom.create!(guest_id: @guest_3.id, room_id: @room_3.id)
+    @guest_rooms_5 = GuestRoom.create!(guest_id: @guest_3.id, room_id: @room_4.id)
+  end
+
+  describe 'When visiting the rooms index page' do
+    it 'has a list of all the rooms including (suite, rate, name of hotel, number of guests)' do 
+      visit "/rooms"
+
+      within("##{@room_1.id}") do
```
cjsim89, line 32: Your tests honestly look super organized and well-written. One bit of advice I have for you when using IDs in HTML is to preface them with some text, and have the ID either in the middle or at the end of the ID string itself. 

This ID is technically something like `#3`; potentially lots of things could have the same ID on a page. A better ID might be `#room_3`. This way you know what kind of object the 3 applies to.

cjsim89 (commented):
Great work overall! Really strong testing habits will take you far.

PR #7, opened 2023-02-21
Tori Enyart

torienyart:
This challenge definitely felt like a step up from the last!  I felt confident in my ability to complete it.... though would like to get more automatic and familiar with migration set-up and forms within views.  I know where to look and where to start with them... but definitely need notes to get through them fully and accurately.

I got through US 3

### DIFF
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -5,3 +5,12 @@
   <p>Nightly Rate: $<%=room.rate%><p>
   <p>Hotel: <%=room.hotel_name%><p>
 <%end%>
+<br>
+<div id="add_stay">
+  <h2>Add a Stay:</h2>
+  <%= form_tag("/guests/#{@guest.id}", method: :get, local: true) do%>
+  <p>Enter Room ID:</p>
+  <%= text_field_tag(:add_room, params[:add_room])%>
+  <%= submit_tag ("Submit")%>
+  <%end%>
+</div>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,1 +1,7 @@
 <h1>Hi, <%=@guest.name%>. Here is your past stay information.</h1>
+<% @rooms.each do |room|%>
+  <h3><%=room.suite%> at the <%=room.hotel_name%></h3>
+  <p>Room Type: <%=room.suite%><p>
+  <p>Nightly Rate: $<%=room.rate%><p>
+  <p>Hotel: <%=room.hotel_name%><p>
+<%end%>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,1 @@
+<h1>Hi, <%=@guest.name%>. Here is your past stay information.</h1>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -2,5 +2,6 @@ Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
   
   get '/guests/:id', to: 'guests#show'
+  get '/rooms', to: 'rooms#index'
 
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,6 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  
+  get '/guests/:id', to: 'guests#show'
+
 end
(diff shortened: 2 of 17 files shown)


app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% @rooms.each do |room|%>
+  <h3><%=room.suite%> at the <%=room.hotel_name%></h3>
+  <p>Room Type: <%=room.suite%><p>
+  <p>Nightly Rate: $<%=room.rate%><p>
+  <p>Hotel: <%=room.hotel_name%><p>
+<%end%>
+<br>
+<div id="add_stay">
+  <h2>Add a Stay:</h2>
+  <%= form_tag("/guests/#{@guest.id}", method: :get, local: true) do%>
```
cjsim89, line 11: FYI - `form_tag` is a deprecated (old) view helper method; we would rather you use `form_with`. We will also be adding to the features of the `form_with` view helper in class soon, so definitely get used to that method instead.

cjsim89: If I were to look at this route separate from the functionality implemented, I would assume this form is responsible for retrieving an existing Guest instance from the database when I hit "Submit". But, that's not what the user story is asking for, and it's not what your form is doing either - it's submitting a Room ID to be related to this existing Guest. 

I think a better route overall would be
```
post "/guest/:id/rooms", to: "guest_rooms#create"
# alternatively: 
post "/guest_rooms", to: "guest_rooms"#create"
```
If you did it the 2nd way I listed above, you'd have to use a hidden field for the `guest_id` in your form, but either would be correct.

spec/features/guest/show_spec.rb
```diff
@@ -0,0 +1,72 @@
+require 'rails_helper'
+
+describe 'As a visitor' do
+  describe 'When I visit a guests show page' do
+    let!(:mtn_1) {Hotel.create!(name: "Sun Valley Main", location: "North Slopes")}
+    let!(:mtn_2) {Hotel.create!(name: "Sun Valley Private", location: "West Slopes")}
+    let!(:tori) {Guest.create!(name: "Tori", nights_stayed: "7")}
+    let!(:penthouse) {tori.rooms.create!(rate: 9998, suite: "Penthouse", hotel_id: mtn_1.id)}
+    let!(:cabin) {tori.rooms.create!(rate: 8798, suite: "Private Cabin", hotel_id: mtn_2.id)}
+    let!(:suite) {tori.rooms.create!(rate: 998, suite: "Suite", hotel_id: mtn_1.id)}
```
cjsim89, line 10: Nice object setup for your tests - generally I suggest creating a few more Guest objects, to play with and make more scenarios with.

spec/features/guest/show_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        expect(page).to have_field("add_room")
+        expect(page).to have_button("Submit")
+      end
+
+      describe 'When I fill in a field with the id of an existing room and click submit' do
+        it 'I am redirected back to the guest show page' do
+          visit "/guests/#{tori.id}"
+
+          within "#add_stay" do
+            fill_in 'add_room', with: "#{suite.id}"
```
cjsim89, line 50: This test doesn't really do what the user story is asking - it's meant to add a Room to this Guest _that isn't already related to this guest_. So if we had an outstanding `Room` object called "3-star" for example, and it wasn't related to the `tori` guest already, we could have an expect line here saying 
`expect(page).to_not have_content("3-star")`

Then, after we subimt the `3_star.id` to this path, we would expect it to appear on the page. 
To do this, we'll need to **create a relationship** between these two objects - the guest that exists, and the room by the ID we're sending in.

cjsim89: This is also why we want to use `post` instead of `get` - POST is for creating new data, and GET is only used to retrieve data for displaying.

config/routes.rb
```diff
@@ -1,3 +1,7 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  
+  get '/guests/:id', to: 'guests#show'
+  get '/rooms', to: 'rooms#index'
+
```
cjsim89, line 6: We'll want a route here for _creating_ a new relationship between Guest and Room, from user story 2: 

```
post "/guests/:id/rooms", to: "guest_rooms#create"
```
This is one way to do it, but we want to make sure the `create` action matches the `post` verb.

PR #8, opened 2023-02-21
Harrison Ryan 2211BE Mid-Mod: symmetrical-guacamole-0156 

hwryan12:
I was able to complete the three user stories. I got stuck on the extension with having the list of guests be unique

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,7 @@
+class GuestRoomsController < ApplicationController
+  def create
+    room = Room.find(params[:room_id])
+    guest_room = GuestRoom.create!(guest_id: params[:id], room: room)
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,5 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+  end
+end
--- a/app/controllers/rooms_controller.rb
+++ b/app/controllers/rooms_controller.rb
@@ -1,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -2,4 +2,8 @@ class Room < ApplicationRecord
    belongs_to :hotel
    has_many :guest_rooms
    has_many :guests, through: :guest_rooms
+
+   def guest_count
+      self.guests.count
+   end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -8,7 +8,6 @@ class Room < ApplicationRecord
    end
 
    def guest_names
-      # binding.pry
       self.guests.pluck(:name)
    end
 end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -6,4 +6,9 @@ class Room < ApplicationRecord
    def guest_count
       self.guests.count
    end
+
+   def guest_names
+      # binding.pry
+      self.guests.pluck(:name)
+   end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,9 @@
+<h1> <%= @guest.name %>'s Page </h1>
+<br>
+Rooms:
+<br>
+<% @guest.rooms.each do |room| %>
+  Rate: $<%= room.rate %>.00, Suite: <%= room.suite %>, Hotel: <%= room.hotel.name %>
+  <br>
+  <%=%>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -7,8 +7,8 @@ Rooms:
   <br>
 <% end %>
 
-<%= form_with url: "guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
-  <%= f.label :suite  %>
-  <%= f.text_field :Suite %>
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label "Suite ID: "  %>
+  <%= f.number_field :room_id %>
   <%= f.submit "Submit" %>
 <% end %>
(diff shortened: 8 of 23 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestRoomsController < ApplicationController
+  def create
+    room = Room.find(params[:room_id])
+    guest_room = GuestRoom.create!(guest_id: params[:id], room: room)
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
```

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,5 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+  end
+end
```

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
...
+   has_many :guests, through: :guest_rooms
+
+   def guest_count
+      self.guests.count
+   end
+
+   def guest_names
+      self.guests.pluck(:name)
+   end
 end
```
abdulredd, line 13: `Then I see a unique list of all guests that have stayed at this hotel`

Just taking a look at that line in the user story, it tells me that we want a list of unique guests for any given hotel. So instead of trying to add a method in the `Room` model, I would be adding it to the `Hotel` model. 

```ruby
  has_many :rooms

  def all_guests
    guests.distinct
  end
end
```

Rails magic FTW. Play around with this and see what SQL gets printed in the terminal. A multi-join!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<% @guest.rooms.each do |room| %>
+  <li> Rate: $<%= room.rate %>.00, Suite: <%= room.suite %>, Hotel: <%= room.hotel.name %></li> 
+  <br>
+<% end %>
+
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label "Suite ID: "  %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit" %>
+<% end %>
```
abdulredd, line 14: I will touch on this in the `routes.rb` file, but the URL for the form isn't correct according to REST.

```ruby
```

```ruby
<%= f.label :room_id,  "Suite ID: " %>
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1> <%= @hotel.name %>'s Page </h1>
+
+<% @hotel.rooms.each do |room| %>
+  <%= room.guest_names %>
+<% end %>
```
abdulredd, line 5: Missing HTML.

`<p><%= room.guest_names %></p>`

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <br>
+  Suite: <%= room.suite %>
+  <br>
+  Rate: $<%= room.rate %>.00
+  <br>
+  Hotel: <%= room.hotel.name %>
+  <br>
+  <%= room.suite %>'s Guest Count: <%= room.guest_count %>
+  <br>
+<% end %>
```
abdulredd, line 12: We mustn't forget about the HTML. Always put rendered ERB in an HTML tag.

```ruby
<p>Suite: <%= room.suite %></p>

<p>Rate: $<%= room.rate %>.00</p>

<p>Hotel: <%= room.hotel.name %></p>

<p><%= room.suite %>'s Guest Count: <%= room.guest_count %></p>
```

config/routes.rb
```diff
@@ -1,3 +1,8 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get "/rooms", to: "rooms#index"
+
+  get "/guests/:id", to: "guests#show"
+  post "/guests/:id/guest_rooms", to: "guest_rooms#create"
+
+  get "/hotels/:id", to: "hotels#show"
```
abdulredd, line 7: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221160432_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221160533_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      redirected back to the guest's show page and I see the room now listed under this guest's rooms" do
+        fill_in :room_id, with: pres.id
+        click_on "Submit"
+        
+        expect(current_path).to eq("/guests/#{jimmy.id}")
+        expect(page).to have_content("Suite: Presidential Room")
+      end
+    end
+  end
+end
```
abdulredd, line 48: 👍 

```ruby
```

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      it "a unique list of all guests that have stayed at this hotel." do
+        # binding.pry
+        save_and_open_page
+        expect(page).to have_content("Jimmy")
+        expect(page).to have_content("Naomi")
+        expect(page).to have_content("Alex")
+      end
+    end
+  end
+end
```
abdulredd, line 33: The way the test is set up it doesn't fully cover what is expected in the user story. Right now you are checking that `"Jimmy"`, `"Naomi"`, or `"Alex"` show up in the browser but not checking if they show up just once. This is what you want:

```ruby
expect(page).to have_content("Jimmy").once
expect(page).to have_content("Naomi").once
expect(page).to have_content("Alex").once
```
or
```ruby
expect(page).to have_content("Jimmy", count: 1)
expect(page).to have_content("Naomi", count: 1)
expect(page).to have_content("Alex", count: 1)
```

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      it "the number of guests that have stayed in that room." do
+  
+        expect(page).to have_content("King Bed Room's Guest Count: 3") 
+        expect(page).to have_content("Twin Bed Room's Guest Count: 1")
+        expect(page).to have_content("Presidential Room's Guest Count: 1")
+      end
+    end
+  end
+end
```
abdulredd, line 44: 👍

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe "relationships" do
+    it {should have_many :guest_rooms}
+    it {should have_many(:rooms).through(:guest_rooms)}
+  end
+end
```

spec/models/room_spec.rb
```diff
@@ -3,5 +3,43 @@
...
+        expect(king.guest_names).to eq(["Jimmy", "Naomi"])
+
+        alex = Guest.create!(name: "Alex", nights: 4)
+        GuestRoom.create!(guest: alex, room: king)
+
+        expect(king.guest_names).to eq(["Jimmy", "Naomi", "Alex"])
+      end
+    end
+  end
+end
```
abdulredd, line 45: 👍

PR #9, opened 2023-02-21
Axel De La Guardia

axeldelaguardia:
Feeling good about this one.

### DIFF
--- a/app/controllers/room_guests_controller.rb
+++ b/app/controllers/room_guests_controller.rb
@@ -1,0 +1,7 @@
+class RoomGuestsController < ApplicationController
+	def create
+		RoomGuest.create!(room_id: params[:room_id], guest_id: params[:guest_id])
+
+		redirect_to "/guests/#{params[:guest_id]}"
+	end
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,16 @@
+<h1><%= @guest.name %></h1>
+
+<ul id='rooms'>
+	<% @rooms.each do |room| %>
+		<li>
+			<%= room.suite %>: <%= room.rate %> - <%= room.hotel.name %>
+		</li>
+	<% end %>
+</ul>
+
+<div id='room_form'>
+	<%= form_with url: "/room_guests/#{@guest.id}", method: :post, local: true do |f| %>
+		<%= f.collection_select :room_id, @all_rooms, :id, :id %>
+		<%= f.submit 'Add Room' %>
+	<% end %>
+</div>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -8,9 +8,9 @@
 	<% end %>
 </ul>
 
-<div id='room_form'>
-	<%= form_with url: "/room_guests/#{@guest.id}", method: :post, local: true do |f| %>
-		<%= f.collection_select :room_id, @all_rooms, :id, :id %>
-		<%= f.submit 'Add Room' %>
-	<% end %>
-</div>
+<%= render partial: 'select_room', locals: {
+	uri: "/room_guests/#{@guest.id}",
+	label: :room_id,
+	list: @all_rooms,
+	button: 'Add Room'
+} %>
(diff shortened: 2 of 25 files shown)


app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,7 @@
+class RoomGuestsController < ApplicationController
+	def create
+		RoomGuest.create!(room_id: params[:room_id], guest_id: params[:guest_id])
```
cjsim89, line 3: Nice job creating this joined controller, it's exactly what I would have done 🙌

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<ul id='rooms'>
+	<% @rooms.each do |room| %>
+		<li>
+			<%= room.suite %>: <%= room.rate %> - <%= room.hotel.name %>
+		</li>
+	<% end %>
+</ul>
+
+<%= render partial: 'select_room', locals: {
+	uri: "/room_guests/#{@guest.id}",
```
cjsim89, line 12: Nice job making this partial! For these ICs it's usually not necessary but this is good practice.

config/routes.rb
```diff
@@ -1,3 +1,10 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+	get '/guests/:id', to: 'guests#show'
+
+	post '/room_guests/:guest_id', to: 'room_guests#create'
```
cjsim89, line 5: I agree with the verb, controller, and action for this route, but usually when we are `POST`-ing to a resource, the ID of that resource will not have been created yet, so it won't have an ID in the URI. 

If you think about the request-response cycle and the rules of RESTful routes, the verb should match with the path and the action in the controller it's being routed to. If we're creating a new resource called `RoomGuest`, then the necessary attributes will be part of that POST data, in the request body. 
Typically, a URI will have an ID that corresponds with its previous string; so an ID on this route would generally be used for a `RoomGuest.id` attribute. Since you're sending in a `guest_id` here, it can easily get confused with a different kind of ID if you just look at the path itself.

So, my suggestion is to put this `guest_id` inside the form body, and keep everything else the same. You'll still have access to it in your params. Alternatively, your route could look like: 
```
post "/guests/:guest_id/rooms", to: "room_guests#create"
```
That way we know we're still creating a relatonship to this room from this guest, and doing it in the place that makes sense, but we are specific about what kind of ID this is.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+	let!(:room_5) { hotel_1.rooms.create(rate: 110, suite: "Double") }
+	let!(:room_3) { hotel_1.rooms.create(rate: 140, suite: "Presidential") }
+	let!(:room_2) { hotel_2.rooms.create(rate: 125, suite: "Twin") }
+	let!(:room_4) { hotel_2.rooms.create(rate: 80, suite: "Studio") }
+	let!(:room_1) { hotel_3.rooms.create(rate: 100, suite: "Deluxe") }
+
+	let!(:guest_1) { Guest.create(name: "Charlize Theron", nights: 2) }
+	let!(:guest_2) { Guest.create(name: "John Doe", nights: 1) }
+	let!(:guest_3) { Guest.create(name: "Mary Jane", nights: 5) }
```
cjsim89, line 16: This data setup is great, you have lots of examples for each type of object. Keep this up.

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+	end
+
+	describe 'list of guests' do
+		it 'it shows a unique list of all guests that have stayed at this hotel' do
+			expect(page).to have_content("List of Guests Who Have Stayed")
+			
+			within '#guests' do
+				expect(page).to have_content(guest_1.name)
+				expect(page).to have_content(guest_2.name)
+				expect(page).to_not have_content(guest_3.name)
```
cjsim89, line 36: These tests are great, well-organized and thorough.

cjsim89 (commented):
You have really strong testing habits, this will serve you very well if you keep it up! Nice job 🎉

PR #10, opened 2023-02-21
Jasmine Hamou

hamouj:
I completed the 3 user stories and the extension.

Based on the feedback I received from the earlier IC, I feel okay about this IC; I have definitely learned to pay close attention to details. Thanks to the multiple joins session on Friday, I have a better understanding of associations and was able to use them to help complete the extension.

I do have a question about correct routing for join table controllers. In User Story 2, I routed the form to "/guests/:id/guest_rooms" with a post method (since I was creating a new association between a room and guest). Is the route path here correct?

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,10 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+
+    GuestRoom.create!(guest: guest, room: room)
+
+    redirect_to "/guests/#{guest.id}"
+  end
+end
--- a/app/models/hotel.rb
+++ b/app/models/hotel.rb
@@ -1,3 +1,8 @@
 class Hotel < ApplicationRecord
-   has_many :rooms
+  has_many :rooms
+  has_many :guests, through: :rooms
+
+  def list_unique_guests
+    self.guests.distinct
+  end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,8 @@
+<h1><%= @guest.name %></h1>
+<% @guest.rooms.each do |room| %>
+<div id="<%= room.suite %>">
+  <h3><li>Suite: <%= room.suite %></li></h3>
+  <p>Hotel: <%= room.hotel.name %></p>
+  <p>Nightly Rate: $<%= room.rate %></p>
+</div>
+<% end %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,13 +1,16 @@
 <h1><%= @guest.name %></h1>
 <% @guest.rooms.each do |room| %>
 <div id="<%= room.suite %>">
+<ul>
   <h3><li>Suite: <%= room.suite %></li></h3>
-  <p>Hotel: <%= room.hotel.name %></p>
   <p>Nightly Rate: $<%= room.rate %></p>
+    <p>Hotel: <%= room.hotel.name %></p>
+</ul>
 </div>
 <% end %>
 <br>
 <div id="add_a_room"> 
+<h3>Add a Room:</h3>
 <%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: :true do |f| %>
   <%= f.label :room_id, "Enter the room ID" %>
   <%= f.number_field :room_id %>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -6,3 +6,11 @@
   <p>Nightly Rate: $<%= room.rate %></p>
 </div>
 <% end %>
+<br>
+<div id="add_a_room"> 
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: :true do |f| %>
+  <%= f.label :room_id, "Enter the room ID" %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit" %>
+<% end %>
+</div>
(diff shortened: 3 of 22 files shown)


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,10 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+
+    GuestRoom.create!(guest: guest, room: room)
```
cjsim89, line 6: Nice job with this controller; it's exactly the same setup that I like to do. 🙌

app/models/hotel.rb
```diff
@@ -1,3 +1,8 @@
 class Hotel < ApplicationRecord
-   has_many :rooms
+  has_many :rooms
+  has_many :guests, through: :rooms
```
cjsim89, line 3: Good job creating this association! Makes the joins a lot easier.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <h3><li>Suite: <%= room.suite %></li></h3>
+  <p>Nightly Rate: $<%= room.rate %></p>
+    <p>Hotel: <%= room.hotel.name %></p>
+</ul>
+</div>
+<% end %>
+<br>
+<div id="add_a_room"> 
+<h3>Add a Room:</h3>
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: :true do |f| %>
```
cjsim89, line 14: I agree with the method and overall setup for this form, I think the URL is not quite RESTful, but it's close... 

When we're writing paths, consider the format of "resource/:which_one". So if we're POST-ing something, that means we're _creating_ something. What thing are we creating here? We're creating a relationship between a Guest and a Room. Do we have a model for that? Yes, it's `GuestRoom`. So, our resource (URL) would be `"guest_rooms"`. Since we're creating this relationship, both of the IDs we need to do it could be contained within the request body, the form data. Since the user is only rqeuired to enter the Room Id, we can make a hidden field for `guest_id` and fill in that value with `@guest.id`. 

Another potential path yo ucould use would be
`url: "/guests/:guest_id/rooms`
and this would still be a POST to the `guest_rooms_controller` `create` action.

cjsim89: The reason I would just do `guest_rooms` for the URL is because that's the resource we're creating. 
The alternative of doing `guests/:guest_id/rooms` is because we're adding a specific room to this specific guest. 

I've seen and used both types of paths for this kind of operation and I think both are equally acceptable.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    
+    @jasmine = Guest.create!(name: "Jasmine", nights: 2)
+    @ivan = Guest.create!(name: "Ivan", nights: 2)
+
+    @penthouse = @hilton.rooms.create!(rate: 350, suite: "Penthouse")
+    @king = @hilton.rooms.create!(rate: 175, suite: "King")
+    @double = @double_tree.rooms.create!(rate: 190, suite: "Double")
+
+    GuestRoom.create!(guest: @jasmine, room: @penthouse)
+    GuestRoom.create!(guest: @jasmine, room: @king)
```
cjsim89, line 16: Your test object setup is good! I like using `before :each` as well.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      expect(current_path).to eq("/guests/#{@jasmine.id}")
+    end
+
+    it "displays the added room after the form is submitted" do
+      visit "/guests/#{@jasmine.id}"
+
+      within "#add_a_room" do
+        fill_in :room_id, with: @double.id
```
cjsim89, line 71: I think one additional expect statement could increase the value of this test overall - before submitting the form, yo ushould expect some attribute of that added object should _not_ appear on the page first. Then, when you expect it to appear later, it carries more weight.

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe 'Extension' do
+    it "displays a unique list of all guests that have stayed at that hotel" do
+      visit "/hotels/#{@hilton.id}"
+
+      within "#guests" do
+        expect(page).to have_content("Name: #{@hailey.name}, Nights: #{@hailey.nights}")
+        expect(page).to have_content("Name: #{@ivan.name}, Nights: #{@ivan.nights}")
+        expect(page).to have_content("Name: #{@jasmine.name}, Nights: #{@jasmine.nights}")
```
cjsim89, line 34: Good test for your extension. If you're looking for uniqueness on a page in a feature test, you should also include a 2nd parameter in `have_content` for `count`. Example: 

```
expect(page).to have_content(@hailey.name, count: 1)
```

Currently, your test is saying this content should be on the page... but it's not necessarily only appearing one time. The `count` parameter makes sure it's only ever appearing once.

spec/models/room_spec.rb
```diff
@@ -2,6 +2,44 @@
...
+
+    describe '#number_of_guests' do
+      it 'returns the number of unique guests for a room' do
+        expect(@penthouse.number_of_guests).to eq(1)
+        expect(@king.number_of_guests).to eq(3)
+        expect(@double.number_of_guests).to eq(2)
+
+        GuestRoom.create!(guest: @jasmine, room: @double)
+        
+        expect(@double.number_of_guests).to eq(3)
```
cjsim89, line 41: Your model tests look great!

cjsim89 (commented):
Great testing habits - both thorough and precise. Keep this up! 🎉

PR #12, opened 2023-02-21
Kassandra Leyba

kassandraleyba:
User Story 2 - I was overthinking quite a bit and it took me longer than I would have liked to get through the stories. I need to spend more time working on practice ICs and reviewing forms. Feedback is appreciated, thanks.


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,8 @@
+class GuestRoomsController < ApplicationController
+  def create
+    gr = GuestRoom.new(room_id: params[:room_id], guest_id: params[:id])
+    # why new instead of create?
+    gr.save
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 8: `# why new instead of create?`

This is a great question. I would almost argue that in this situation, `create` would be the better choice. Personally I use `new`/`save` when I am also adding a sad path for flash messages. Example:

```ruby
def create
  gr = GuestRoom.new(room_id: params[:room_id], guest_id: params[:id])
  
  if gr.save # if it saves successfully, do this
    redirect_to "/guests/#{params[:id]}"
  else # otherwise do this
    flash[:error] = gr.errors.full_messages
    render :new
  end
end
```

If we just used `create` we wouldn't be able to use branching with `if`/`else`. Please let me know if you want to discuss this further.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,5 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+  end
+end
```

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
 end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <% end %> 
+</ul>
+
+<h3>Guest Form:</h3>
+
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label "ID: " %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit" %>
+<% end %>
```
abdulredd, line 22: I will touch on this in the `routes.rb` file, but the URL for the form isn't correct according to REST.

```ruby
```

```ruby
<%= f.label :room_id,  "ID: " %>
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Hotels Show Page</h1>
+
+<p>Hotel Name: <%= @hotel.name %></p>
+<p>Guests: <% @hotel.rooms.each do |room| %> 
+  <% room.guests.each do |guest| %>
+    <p><%= guest.name %></p>
+  <% end %>
+<% end %>
```
abdulredd, line 8: Great job attempting the extension. However, this doesn't quite do what the extension is asking. 

The extension wants a unique list of all guests that have stayed at this hotel. With the way this is implemented is a single guest stayed in different rooms at the hotel, there's potential for this guest's name to show up more than once.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Room Index Page</h1>
+
+<% @rooms.each do |room| %>
+  <p>Room: <%= room.suite %></p>
+  <p>Rate: <%= room.rate %></p>
+  <p>Hotel Name: <%= room.hotel.name %></p>
+  <p>Number of Guests: <%= room.guests.count %></p>
+<% end %>
```
abdulredd, line 8: If we think about line `7` where you are doing `room.guests.count` and think about what it's doing. On that line, it's taking `room.guests` and then calling the ActiveRecord aggregate method `count` on it. This is performing a database call, and because this is hitting the database, this code needs to be in the `Room` model.

I just want to compare that to line `6`, where you are doing `room.hotel.name`. On this line, you aren't calling the database because when we called on all the `@rooms` in the controller, those rooms, came with the hotel data that is linked to each room, and each of those hotels came with its attributes, including the `name`. No additional database call is being made there.

config/routes.rb
```diff
@@ -1,3 +1,9 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/guests/:id', to: 'guests#show'
+  post '/guests/:id/guest_rooms', to: 'guest_rooms#create'
+
+  get '/rooms', to: 'rooms#index'
+
+  get '/hotels/:id', to: 'hotels#show'
```
abdulredd, line 8: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221161158_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221161725_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.belongs_to :guest, foreign_key: true
+      t.belongs_to :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      fill_in :room_id, with: penthouse.id
+      click_on "Submit"
+      # save_and_open_page
+
+      expect(current_path).to eq("/guests/#{charlize.id}")
+      expect(page).to have_field(:room_id)
+      expect(page).to have_button("Submit")
+    end
+  end
+end
```
abdulredd, line 50: In your user story test, after filling out the form and submitting it, there should be additional assertions that check that the actual element was added to the page successfully.

```ruby
expect(page).to have_content("Room: Penthouse")
expect(page).to have_content("Rate: 175")
expect(page).to have_content("Hotel Name: Everest View Hotel")
```
 
---

```ruby
```

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      expect(page).to  have_content("Hotels Show Page\n")
+      expect(page).to have_content("Hotel Name: Everest View Hotel\n")
+      expect(page).to have_content("Guests:\nScarlett Johansson\nKeanu Reeves")
+
+      expect(page).to_not have_content("Hotel Name: Echo Mountain Inn")
+      expect(page).to_not have_content("Guests: Charlize Theron")
+      save_and_open_page
+    end
+  end
+end
```
abdulredd, line 34: I spoke a little about this in the view, but in order to have properly tested the functionality, you would have needed to have a guest stay in multiple rooms at the same hotel in order to check that they only show up once. This would have also forced you to create a `Hotel` model method to pick unique guests which was the intention behind this user story.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(page).to have_content("Number of Guests: 1")
+
+      expect(page).to have_content("Room: Penthouse")
+      expect(page).to have_content("Rate: 175")
+      expect(page).to have_content("Hotel Name: Everest View Hotel")
+      expect(page).to have_content("Number of Guests: 1")
+      # save_and_open_page
+    end
+  end
+end
```
abdulredd, line 37: 👍

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  describe 'relationships' do
+    it {should belong_to :guest}
+    it {should belong_to :room}
+  end
+end
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it {should have_many :rooms}
+    it {should have_many :guest_rooms}
+    it {should have_many(:rooms).through(:guest_rooms)}
+  end
+end
```
abdulredd, line 9: 👍

spec/models/room_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Room, type: :model do
   describe 'relationships' do
     it {should belong_to :hotel}
+    it {should have_many :guest_rooms}
+    it {should have_many(:guests).through(:guest_rooms)}
   end
 end
```
abdulredd, line 9: Having moved the `count` method to the model, you would have also needed to test it here.

PR #14, opened 2023-02-21
Weston Sandfort

sandfortw:


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+    GuestRoom.create(guest_id: params[:id], room_id: params[:room_id])
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  # def update
+  #   GuestRoom.create(guest_id: params[:id], room_id: params[:room_id])
+  #   redirect_to "/guests/#{params[:id]}"
+  # end
+  
+end
```
abdulredd, line 12: 👍

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+
+  def show
+    @hotel = Hotel.find(params[:id])
+  end
+end
```
abdulredd, line 6: 👍

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RoomsController < ApplicationController
+
+  def index
+    @rooms = Room.all
+  end
+end
```
abdulredd, line 6: 👍

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
    has_many :rooms
+
+   def guests
+      Guest.joins(:guest_rooms).joins(:rooms).where("rooms.hotel_id = #{self.id}").distinct
+   end
 end
```
abdulredd, line 7: I really the `guests` method here. What if I told you there was an easier way to do the exact same logic via an association?

```ruby
  has_many :rooms

  def all_guests
    guests.distinct
  end
end
```

Rails magic FTW. Play around with this and see what SQL gets printed in the terminal. A multi-join just like you defined by hand in your own logic!

app/models/room.rb
```diff
@@ -1,3 +1,13 @@
...
+   has_many :guests, through: :guest_rooms
+
+   def hotel_name
+     hotel.name
+   end
+
+   def number_of_guests
+     guests.distinct.count
+   end
 end
```
abdulredd, line 13: 👍 

I don't mind the `hotel_name` method, but you don't necessarily need to make it a separate method either. The logic, `hotel.name`, isn't really making any inquiry to the database and is just pulling the name from the hotel object. So you could have just used that in the view.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p>Rate: <%= room.rate %> </p>
+  <p>Hotel: <%= room.hotel_name %> </p>
+<% end %>
+
+<h2> Add a Room by its ID: </h2>
+<%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |form|  %>
+  <%= form.label :room_id %>
+  <%= form.number_field :room_id %>
+  <%= form.submit :Submit %>
+<% end %>
```
abdulredd, line 16: 🙌 

Great job setting the form to the correct path!

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1> <%= @hotel.name %> Show Page </h1>
+
+<h2> List of Guests </h2>
+<% @hotel.guests.each do |guest| %>
+  <p> <%= guest.name %> </p>
+<%end %>
```
abdulredd, line 6: 👍

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1> Rooms Index</h1>
+
+<% @rooms.each do |room| %>
+  <h3> Suite: <%= room.suite %> </h3>
+  <p> Rate: <%= room.rate %> </p>
+  <p> Hotel: <%= room.hotel_name %> </p>
+  <p> Number of Guests: <%= room.number_of_guests %> </p>
+<% end %>
```
abdulredd, line 8: 👍

config/routes.rb
```diff
@@ -1,3 +1,7 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get "/guests/:id", to: "guests#show"
+  # patch "/guests/:id", to: "guests#update"
+  post "/guests/:id/rooms", to: "guest_rooms#create"
+  get "/rooms", to: "rooms#index"
+  get "/hotels/:id", to: "hotels#show"
```
abdulredd, line 6: 👍

db/migrate/20230221160350_create_guests.rb
```diff
@@ -0,0 +1,9 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 9: 👍

db/migrate/20230221160648_create_guest_rooms.rb
```diff
@@ -0,0 +1,9 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.belongs_to :guest
+      t.belongs_to :room
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 9: 👍

spec/features/guest/show_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    expect(page).to_not have_content("Suite: #{room3.suite}")
+    expect(page).to_not have_content("Rate: #{room3.rate}")
+    fill_in 'room_id', with: room3.id
+    click_on 'Submit'
+    save_and_open_page
+    expect(current_path).to eq("/guests/#{charlize.id}")
+    expect(page).to have_content("Suite: #{room3.suite}")
+    expect(page).to have_content("Rate: #{room3.rate}")
+  end
+end
```
abdulredd, line 55: I know that during an IC, time is of the essence but I would like to see some better test set up. `describe 'guest show page'` is okay but I would like to see some additional `describe` blocks being used to set up your tests.

`it 'user story 1'` doesn't tell me anything about the test. Even though you have the user stories commented in those `it` blocks, if I were to run the tests with documentation output, the output at first glance wouldn't tell me much more than there is a `user story 1` and a `user story 2`.

Beyond that, the tests are set up perfectly, and I like how you used string interpolation in your assertions to give them a little more weight and less ambiguity.

Only recommendation other recommendation I would make is to start working in `within` blocks into your tests.

spec/features/hotel/show_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    guest_room4 = GuestRoom.create!(room_id: room1.id, guest_id: charlize.id)
+
+    visit "/hotels/#{budapest.id}"
+    expect(page).to have_content("#{budapest.name} Show Page")
+    expect(page).to have_content("List of Guests")
+    expect(page).to have_content(charlize.name, count: 1)
+    expect(page).to have_content(danny.name)
+    expect(page).to have_content(mark.name)
+  end
+end
```
abdulredd, line 27: I know that during an IC, time is of the essence but I would like to see some better test set up. `describe 'extension'` is not descriptive of what this feature test is for, this is a test for the hotel show page, and should be in the `describe`. I would also like to see some additional `describe` blocks being used to set up your tests.

`it 'has a list of all unique guests'` is a better description of what is being expected in the test compared to your guests show spec.

Beyond that, the tests are set up perfectly, and I like how you used string interpolation in your assertions to give them a little more weight and less ambiguity.

The only recommendation other recommendation I would make is to start working in `within` blocks into your tests.

spec/features/room/index_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      expect(page).to have_content(room1.rate)
+      expect(page).to have_content(room1.hotel_name)
+      expect(page).to have_content(room1.number_of_guests)
+
+      expect(page).to have_content(room2.suite)
+      expect(page).to have_content(room2.rate)
+      expect(page).to have_content(room2.number_of_guests)
+    end
+  end
+end
```
abdulredd, line 31: Same comments I've made previously about better test organization with `describe` and `within`.

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  describe 'relationships' do
+    it {should belong_to :guest}
+    it {should belong_to :room}
+  end
+end
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it '#name' do
+    charlize = Guest.create!(name:"Charlize Theron", nights: 3)
+    expect(charlize.name).to eq("Charlize Theron")
+  end
+
+  it '#nights' do
+    charlize = Guest.create!(name:"Charlize Theron", nights: 3)
+    expect(charlize.nights).to eq(3)
+  end
+end
```
abdulredd, line 18: You do not need to test for attributes on your Models. We can safely assume that ActiveRecord is making those available for us.

spec/models/hotel_spec.rb
```diff
@@ -4,4 +4,27 @@
...
+    charlize = Guest.create!(name:"Charlize Theron", nights: 3)
+    danny = Guest.create!(name:"Danny", nights: 2)
+    mark = Guest.create!(name:"Mark", nights: 3) 
+    guest_room1 = GuestRoom.create!(room_id: room1.id, guest_id: charlize.id)
+    guest_room2 = GuestRoom.create!(room_id: room1.id, guest_id: danny.id)
+    guest_room3 = GuestRoom.create!(room_id: room1.id, guest_id: mark.id)
+    guest_room4 = GuestRoom.create!(room_id: room1.id, guest_id: charlize.id)
+    expect(budapest.guests.sort).to eq([charlize, danny, mark].sort)
+  end
 end
```
abdulredd, line 30: You do not need to test for attributes on your Models. We can safely assume that ActiveRecord is making those available for us.

Great job on your `#guests` test. I'm wondering why you have to sort both arrays. Would something like `.to contain_exactly()` made it so you wouldn't have to sort?

spec/models/room_spec.rb
```diff
@@ -3,5 +3,35 @@
...
+  it '#number_of_guests' do 
+    budapest = Hotel.create!(name: "Grand Budapest", location: 'Turkey')
+    room1 = Room.create!(hotel_id: budapest.id, suite: "Presidential", rate: 125)
+    charlize = Guest.create!(name:"Charlize Theron", nights: 3)
+    manny = Guest.create!(name:"Manny", nights: 2)
+    guest_room1 = GuestRoom.create!(room_id: room1.id, guest_id: charlize.id)
+    guest_room2 = GuestRoom.create!(room_id: room1.id, guest_id: manny.id)
+    expect(room1.number_of_guests).to eq(2)
   end
 end
```
abdulredd, line 37: You do not need to test for attributes on your Models. We can safely assume that ActiveRecord is making those available for us.

spec/models/hotel_spec.rb
sandfortw: Thanks for that, Abdul. I just forgot how to compare two unsorted arrays. `.to contain exactly()` sounds like it will solve this problem going forward. I have some tests in my current project that are solving this problem the same way, so I'll make that change.

PR #15, opened 2023-02-21
Melony Franchini - Mid-Mod IC

MelTravelz:
**Reflection on how you felt you did with this challenge:** 
_I'm really frustrated I spent nearly an hour trying to sort out my association error message... in the end I had an 's' on the file name and NOTHING was wrong with my set up except that. That really got my mind in the wrong space and the rest of the time I was very nervous. However, I was able to finish all user stories and even add validations and a flash error message for the form. I know if I hadn't gotten off to a bad start I would have had time to think through the extension more and probably finish it too... I think I needed to possibly be an instance method after all. Anyways, I'm kinda upset about this all, but in the end I made it work._

**What story you got through:** 
_Finished User Story 3, made it partially through extension but ran out of time_


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,15 @@
+class GuestRoomsController < ApplicationController
+
+  # def create
+  #   GuestRoom.create!(room_id: params[:room_id], guest_id: params[:id])
+  #   redirect_to "/guests/#{params[:id]}"
+  # end
+
+  def create
+    gr = GuestRoom.new(room_id: params[:room_id], guest_id: params[:id])
```
cjsim89, line 9: Great work with this `GuestRoomsController`, it's exactly the kind of setup I would do. 🙌

app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+      if !gr.save
+        flash[:alert] = gr.errors.full_messages.join(', ')
```
cjsim89, line 11: Just so you're aware, the user story mentions this not being necessary for the IC - and in the future we won't ask you to implement a sad path in an IC or final. But, this is good practice!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<br>
+<h4>Past Stays:</h4>
+<ul>
+  <% @guest.rooms.each do |room| %>
+    <li> <%= room.suite %> - rate: $<%= room.rate %>/night, hotel: <%= room.hotel.name %></li>
+    <br>
+  <% end %>
+</ul>
+  
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |form|  %>
```
cjsim89, line 11: I think the form body is good and the method is also good, but I think the URL of this form could be more RESTful. 

When we're writing paths, consider the format of "resource/:which_one". So if we're POST-ing something, that means we're creating something. I like to ask myself the following kinds of questions: 
- What thing are we creating here? We're creating a relationship between a Guest and a Room. 
- Do we have a model for that? Yes, it's `GuestRoom`. So, our resource (URL) would be `"guest_rooms"`. 

Since we're creating this relationship, both of the IDs we need to do it could be contained within the request body, the form data. Since the user is only rqeuired to enter the Room Id, we can make a hidden field for `guest_id` and fill in that value with `@guest.id`.

Another potential path you could use would be
`url: "/guests/:guest_id/rooms`
and this would still be a POST to the `guest_rooms_controller#create` action.

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1> <%= @hotel.name %>'s Page</h1>
+<br>
+<h4>All Guests who've stayed with us:</h4>
+<%# <ul>
+  <% @hotel.rooms.each do |room| %>
```
cjsim89, line 5: This looks like a good start to the extension! Feel free to finish this out for good practice.

app/views/guests/show.html.erb
cjsim89: To be clear, the reason I would just do guest_rooms for the URL is because that's the resource we're creating.
The alternative of doing `guests/:guest_id/rooms` is because we're adding a specific room to this specific guest.

I've seen and used both types of paths for this kind of operation and I think both are equally acceptable.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      @room3 = Room.create!(rate: 200, suite: "Loft", hotel_id: @hotel2.id)
+      @room4 = Room.create!(rate: 450, suite: "Honeymoon", hotel_id: @hotel2.id)
+
+      @guest1 = Guest.create!(name: "Letti")
+
+      GuestRoom.create!(room: @room1, guest: @guest1)
+      GuestRoom.create!(room: @room2, guest: @guest1)
+      GuestRoom.create!(room: @room3, guest: @guest1)
+
+      visit "/guests/#{@guest1.id}"
```
cjsim89, line 20: Nice job with your object setup in your tests! 

One thing to be careful of in the future - if in any `it` blocks underneath this `before do` _alters_ any data or creates more relationships within the `it` blocks themselves, the page will have already been visited, and won't see that data updated. So, sometimes it is better to do the `visit...` line inside each `it` block, rather than in a `before :each`. Just something to keep in mind.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      Guest.create!(name: "Calum")
+      Guest.create!(name: "Nika")
+    end
+
+    it "::total_count" do
+      expect(Guest.total_count).to eq(3)
+
+      Guest.create!(name: "Melony")
+
+      expect(Guest.total_count).to eq(4)
```
cjsim89, line 25: So, this is a tricky method to test... You implemented`total_count` as a Class method on Guest. This is fine, but when we do this, we have 2 different ways we can call the method --- on a collection of ALL the guests, or a collection of SOME of the guests based on an associated Room. The 2nd way is what the user story asks for, and it's how you have it implemented in your View and Feature tests, which is great.

However, we also need to be sure we're testing for that in the model test - the model tests are unit/atomic tests that ensure the functionality we designed is working the way it's intended. This test is looking at calling on ALL the guests, we just need to have additional tests to call on specific associations of Rooms, which is how you're using the method in the view/elsewhere.

So, if this included something like: 
```
room1 = Room.create(...)
room2 = Room.create(...)

GuestRoom.create(guest: mel, room: room1)
GuestRoom.create(guest: calum, room: room1)

```
then that would be a complete test. As it stands now though, I would say this model is not completely tested.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "::uniq_guest_names" do
+      expect(Guest.uniq_guest_names).to eq(["Nika", "Letti", "Calum"])
+
+      Guest.create!(name: "Nika")
+
+      expect(Guest.uniq_guest_names).to eq(["Nika", "Letti", "Calum"])
+
+      Guest.create!(name: "Melony")
+
+      expect(Guest.uniq_guest_names).to eq(["Nika", "Letti", "Calum", "Melony"])
```
cjsim89, line 37: Similar here to above; the extension asks for this list to be all the guests _that stayed at this Hotel_. I think a more straightforward way of implementing this would be to make an instance method on Hotel that gets all of its related `guests`, maybe through an association.... Or, you could write a join starting in Hotel and ending up in Guests. 

Also, this test doesn't really cover "uniqueness" since the guests are just created and then listed. The "uniqueness" of each Guest in this scenario really comes from a relationship between Hotel and Guest, through the GuestRoom join table. In other words, even if a Guest stayed a Hotel 2-3-4 times, they should only appear in this list _once_.

cjsim89 (commented):
Great work overall, and nice job getting to the extension! Feel free to finish it out and resubmit for additional feedback if you like.

PR #16, opened 2023-02-21
Brian Mid Mod IC

Bphayes1200:


app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Rooms</h1>
+<% @rooms.each do |room| %>
+  <h4><%= room.suite %></h4>
+  <%= "Rate: #{room.rate}" %><br>
+  <%= "hotel: #{room.hotel.name}" %><br>
+  <%= "Number of total guests: #{room.guests.count}" %>
```
cjsim89, line 6: Be careful here - technically `.count` in this case is an ActiveRecord method that changes the SQL query underneath. So, we want this in the `Room` model inside of a method.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <%= room.suite %><br>
+    <ul> 
+      <li><%= room.rate %></li>
+      <li><%= room.hotel.name %></li>
+    </ul>
+  <% end %>
+</div>
+<div class="form">
+<b>Add a New Room</b>
+ <%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |form| %>
```
cjsim89, line 13: Good form setup!

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,12 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+
+  def create
+    room_guest = RoomGuest.create!(room_id: params[:room_id], guest_id: params[:id] )
```
cjsim89, line 9: The functionality (the "meat") of this method is fine; I don't think it necessarily belongs inside the `GuestsController`.... Consider if we needed to create a new `Guest` object; where would this go? 

When you're trying to figure out where to put a specific controller action, think about what is being done in that controller action first - in here, we're creating a new `RoomGuest`. We're not creating a new `Guest`, just using an ID to make a new _relationship_ between that Guest and a specific Room. So, we can make a new controller called `room_guests_controller` and put this `create` action in there.

config/routes.rb
```diff
@@ -1,3 +1,7 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  post '/guests/:id/rooms', to: 'guests#create'
```
cjsim89, line 3: Overal this route is fine, but see my above comment on which controller this should be routed to.

spec/models/hotels_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      hotel = Hotel.create!(name: "Marriott", location: "Denver")
+      guest_1 = Guest.create!(name: "Brian", nights: 2)
+      guest_2 = Guest.create!(name: "Brian", nights: 2)
+      room_1 = Room.create!(rate: 100, suite: "President suite", hotel_id: hotel.id)
+      room_2 = Room.create!(rate: 150, suite: "Honeymoon Suite", hotel_id: hotel.id)
+      RoomGuest.create!(room: room_1, guest: guest_1)
+      RoomGuest.create!(room: room_2, guest: guest_2 )
+      RoomGuest.create!(room: room_2, guest: guest_1)
+
+      expect(hotel.unique_guests).to eq([guest_1, guest_2])
```
cjsim89, line 15: Nice test here! I think the only thing we're missing is the `Hotel` model itself. I would try to make the models that are outlined in the README _before_ starting the user stories in the future. But, feel free to work on that and resubmit for more feedback.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+          expect(page).to have_no_content(room_3.suite)
+          expect(page).to have_no_content(room_3.rate)
+        }
+
+        fill_in :room_id, with: room_3.id
+        click_button 'Submit'
+
+        expect(page).to have_current_path("/guests/#{guest.id}")
+
+        within('.rooms') {
```
cjsim89, line 67: You're using lots of `within` blocks and they're all used well - keep this up!

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        room_2 = Room.create!(rate: 150, suite: "Honeymoon Suite", hotel_id: hotel.id)
+        room_3 = Room.create!(rate: 1000, suite: "Space Suite", hotel_id: hotel.id)
+        RoomGuest.create!(room: room_1, guest: guest )
+        RoomGuest.create!(room: room_2, guest: guest )
+
+        visit "/guests/#{guest.id}"
+
+        within('.rooms') {
+          expect(page).to have_no_content(room_3.suite)
+          expect(page).to have_no_content(room_3.rate)
```
cjsim89, line 59: Good job testing for the "negative" case; it gives more weight to the test below, when the content is expected to be there.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'rails_helper'
+
+RSpec.describe 'Guest Show' do 
+  describe 'as a visitor' do 
+    describe 'when I visit a guest show page' do 
+      it 'will have the guest name and a list of all rooms they have stayed in' do
```
cjsim89, line 6: You have great object setup in each test; you may want to consider using a `before :each` block with instance variables to cut down on the amount of typing you might be doing; not required just an idea!

PR #17, opened 2023-02-21
Dawson Timmons Mid-Mod Submission

DMTimmons1:
This submission is for Dawson Timmons and completes up to and including User Story 3.

I think I did pretty decent on this Mid-Mod. The only thing i'm really worried about is if I should've had the count of all the guests who have stayed in a room as a variable in the model, then to be called in my view. Instead I just did the count of room_guests for that specific iteration in the view, which made more sense to me.

As mentioned above, all user stories have been completed, the extension has not been completed.


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
```

app/controllers/room_guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RoomGuestsController < ApplicationController
+  def create
+    RoomGuest.create!(room_id: params[:room_id], guest_id: params[:id])
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 6: 👍

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :room_guests
+  has_many :rooms, through: :room_guests
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :room_guests
+   has_many :guests, through: :room_guests
 end
```
abdulredd, line 5: All the associations are set up correctly. 

I want to take the opportunity to recommend that the `guests.count` that you're currently doing in the view actually be made into a model method in the `Room` model. The way it's set up, you're actually performing ActiveRecord in your view which breaks MVC convention.

app/models/room_guest.rb
```diff
@@ -0,0 +1,4 @@
+class RoomGuest < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <%= form.submit "Submit"%>
+<% end %>
+<h3> <%=@guest.name%>'s Rooms: </h3>
+
+<% @guest.rooms.each do |room| %>
+  <p> Suite: <%= room.suite %> </p>
+  <p> Nightly Rate: <%= room.rate %> </p>
+  <p> Hotel: <%= room.hotel.name %> </p>
+  <br>
+<% end %>
```

```ruby
```

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+
+<% @rooms.each do |room| %>
+  <h3> Suite: <%= room.suite %> </h3>
+  <h3> Nightly Rate: <%= room.rate %> </h3>
+  <h3> Hotel: <%= room.hotel.name %> </h3>
+  <div id = "count-<%= room.id %>">
+  <h3> Total Number of Guests: <%= room.room_guests.count %> </h3>
+  </div>
+  <br>
+<% end %>
```
abdulredd, line 11: I mentioned this already in your `Room` model, but by doing `<%= room.room_guests.count %>`, we are actually performing ActiveRecord in the view, thus breaking MVC conventions. You want to move this to the `Room` model.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/guests/:id', to: 'guests#show'
+  post "/guests/:id/room_guests", to: "room_guests#create"
+  get "/rooms", to: "rooms#index"
```
abdulredd, line 4: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221161247_create_guests.rb
```diff
@@ -0,0 +1,9 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 9: 👍

db/migrate/20230221161523_create_room_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRoomGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :room_guests do |t|
+      t.belongs_to :room, foreign_key: true
+      t.belongs_to :guest, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      fill_in :room_id, with: "#{room_3.id}"
+      click_button "Submit"
+
+      expect(current_path).to eq("/guests/#{guest_1.id}")
+      expect(page).to have_content("Penthouse")
+      expect(page).to have_content(233)
+      expect(page).to have_content("Stanley Hotel")
+    end
+  end
+end
```
abdulredd, line 77: There are essentially 2 good ways to add additional weight to a test, 1 would be to use within blocks to ensure that you are checking the content in only a certain section of the page, and the second would be to use some interpolation or at least write out completely what you expect to see in the page.

Right now, as written, `expect(page).to have_content(145)` would return true for any 145 on the page, even it was part of a phone number or something like that. Just looking at the assertion, you can't tell what `145` is referring to.

If you used a `within` block, it would be easier to identify the context.

Using interpolation also works as well. In your view, you have `<p> Nightly Rate: <%= room.rate %> </p>` so I would mirror my test assertion similarly.

```ruby
expect(page).to have_content("Nightly Rate: #{room.rate}")
```
or
```ruby
expect(page).to have_content("Nightly Rate: 145")
```

This goes for the other fields as well. This way we've made it obvious what `145` is for.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect(page).to have_content(2)
+      end
+
+      within "#count-#{room_3.id}" do
+        expect(page).to have_content(1)
+      end
+
+    end
+  end
+end
```
abdulredd, line 56: Same comments that I made in the `guests/show_spec.rb` about trying to add additional weight to your tests.

Great job working in some `within` blocks here.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it { should have_many :room_guests }
+    it { should have_many(:rooms).through(:room_guests) }
+  end
+end
```

spec/models/room_guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe RoomGuest, type: :model do
+  describe 'relationships' do
+    it { should belong_to :room }
+    it { should belong_to :guest }
+  end
+end
```
abdulredd, line 8: 👍

spec/models/room_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe Room, type: :model do
   describe 'relationships' do
-    it {should belong_to :hotel}
+    it { should belong_to :hotel }
+    it { should have_many :room_guests }
+    it { should have_many(:guests).through(:room_guests) }
   end
 end
```

PR #18, opened 2023-02-21
Caleb Thomas

cjthomas00:
This was a good test. I felt confident in my ability to perform all of the user stories. I tried doing within testing for the extension, and only happy path tested. I tried sad paths, but I wasn't writing the expects correctly. I was running out of time, but otherwise I would have looked up how to write the test more thorough. 

The only line of code not tested is in my features. It is the error messages in the #create for GuestRooms Controller. 

I made it through the extension.


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,10 @@
+class GuestRoomsController < ApplicationController
+  def create
+    gc = GuestRoom.new(room_id: params[:room_id], guest_id: params[:id]) 
+
+    if !gc.save
+    flash[:alert] = gc.errors.full_messages
+    end
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 10: 👍 

I like that you included a sad path check here in your `create` action.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GuestsController < ApplicationController
+  def show 
+    @guest = Guest.find(params[:id])
+  end
+end
```

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,5 @@
+class HotelsController < ApplicationController
+  def show 
+    @hotel = Hotel.find(params[:id])
+  end
+end
```

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,17 @@
...
+   end
+
+   def guest_count
+      guests.distinct.count
+   end
+
+   def unique_guest_name
+      guests.distinct
+   end
 end
```
abdulredd, line 17: 👍 

I don't mind the `hotel_name` method, but you don't necessarily need to make it a separate method either. The logic, `hotel.name`, isn't really making any inquiry to the database and is just pulling the name from the hotel object. So you could have just used that in the view.

Actually, after taking a look at your view, it looks like you didn't use this method at all, and could have probably gotten rid of it.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <li>Hotel: <%= room.hotel.name%></li>
+  <br>
+<% end %>
+</ul>
+
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f|  %>
+ <%= f.label "ID: " %>
+ <%= f.number_field :room_id %>
+ <%= f.submit "Submit" %>
+<% end %>
```

```ruby
```

---

Also, the `f.label` should also have a name attribute that is the same as the field that the label is being used for.

```ruby
```

---

Last thing, I already mentioned this in your `Room` model but, you defined a `hotel_name` method but aren't using it.

config/routes.rb
```diff
@@ -1,3 +1,9 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-end
+  get '/guests/:id', to: "guests#show"
+  post '/guests/:id/guest_rooms', to: "guest_rooms#create"
+
+  get '/rooms', to: "rooms#index"
+
+  get '/hotels/:id', to: "hotels#show"
+ end
```
abdulredd, line 9: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<% @hotel.rooms.each do |room|%>
+<div id='room_info-<%=room.id%>'>
+  <h3>Suite: <%= room.suite %></h3> 
+  <% room.unique_guest_name.each do |guest|%>
+     <li>Guest: <%= guest.name%></li> 
+      <%end%> 
+    <%end%>
+</ul>
+</div>
+
```
abdulredd, line 14: 👍

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+<ul>
+<br>
+<%@rooms.each do |room|%>
+  <li>Room Suite: <%= room.suite %></li>
+  <li>Nightly Rate: <%= room.rate%></li>
+  <li>Hotel: <%= room.hotel.name%></li>
+  <li># of Guests That Have Stayed in Room: <%= room.guest_count%></li>
+  <br>
+<% end %>
+</ul>
```
abdulredd, line 11: 👍

db/migrate/20230221161144_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221161427_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/seeds.rb
```diff
@@ -5,3 +5,26 @@
...
+GuestRoom.create!(guest:@matt, room: @room1)
+GuestRoom.create!(guest:@matt, room: @room2)
+GuestRoom.create!(guest:@mike, room: @room2)
+GuestRoom.create!(guest:@mike, room: @room1)
+GuestRoom.create!(guest:@mike, room: @room4)
+GuestRoom.create!(guest:@mike, room: @room5)
+GuestRoom.create!(guest:@matt, room: @room5)
+GuestRoom.create!(guest:@matt, room: @room5)
+GuestRoom.create!(guest:@mark, room: @room6)
+GuestRoom.create!(guest:@mike, room: @room6)
```
abdulredd, line 30: When you're working through and IC, it really isn't required to have to create any seed data since you will be working in the test files for the most part. If you wanted to see what the page looked like in the browser, you would just use `save_and_open_page`

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        fill_in :room_id, with: "#{@room3.id}"
+        click_on "Submit"
+
+        expect(current_path).to eq("/guests/#{@mike.id}")
+        expect(page).to have_content("Room Name: Honeymoon")
+        expect(page).to have_content("Rate: 250")
+      end
+    end
+  end
+end
```
abdulredd, line 45: 👍 

```ruby
```

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        expect(page).to have_content("Suite: #{@room2.suite}")
+        expect(page).to have_content("Guest: #{@mike.name}")
+        expect(page).to have_content("Guest: #{@matt.name}")
+        end
+        expect(page).to have_content("Suite: #{@room3.suite}")
+      end
+    end
+  end
+
+end
```
abdulredd, line 38: 👍 

Great job working in those `within` blocks into your test. Don't forget to remove the `save_and_open_page` part when you are done inspecting the test.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        expect(page).to have_content("Hotel: #{@room2.hotel.name}")
+        expect(page).to have_content("# of Guests That Have Stayed in Room: #{@room2.guest_count}")
+        expect(page).to have_content("Room Suite: #{@room3.suite}")
+        expect(page).to have_content("Nightly Rate: #{@room3.rate}")
+        expect(page).to have_content("Hotel: #{@room3.hotel.name}")
+        expect(page).to have_content("# of Guests That Have Stayed in Room: #{@room3.guest_count}")
+      end
+    end
+  end
+end
```
abdulredd, line 42: 👍

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  describe "relationships" do
+    it { should belong_to(:room) }
+    it { should belong_to(:guest) }
+  end
+end
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe "relationships" do
+    it { should have_many(:guest_rooms) }
+    it { should have_many(:rooms).through(:guest_rooms) }
+  end
+end
```

spec/models/room_spec.rb
```diff
@@ -3,5 +3,38 @@
...
+        expect(@room2.guest_count).to_not eq(nil)
+      end
+
+      it '#unique_guest_name' do
+        expect(@room1.unique_guest_name).to eq([@mike])
+        expect(@room2.unique_guest_name).to eq([@mike,@matt])
+      end
+    end
   end
 end
```
abdulredd, line 40: 👍

PR #19, opened 2023-02-21
Bobby Luly- Mid Mod

Bobsters986:
- I feel as if I did pretty well on the setup and User Stories 1-3. I ran out of time trying to write the Active Record/joins for the extension
- I got through User Story 3, got through some of the extension


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+    gr = GuestRoom.new(room: room, guest_id: params[:id])
+
+    if gr.save
+      flash[:success] = "Room created!"
+      redirect_to "/guests/#{params[:id]}"
+    else
+      flash[:error] = "Room creation failed: #{gr.errors.full_messages.join}"
+    end
+  end
+end
```
abdulredd, line 13: 👍 

Great job going the extra mile here and considering the sad path. The only thing is what is the controller to do if the `else` is hit? There is no redirect or render, so the action will attempt to render a page called `create.html.erb` which doesn't exist, and then call the request to error out.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @guests = @hotel.guest_names
+  end
+end
```
abdulredd, line 6: I know you were currently working through the extension before running out of time so I won't dwell on this much but you have `::guest_names` defined as a class method in the `Hotel` model, but here you are using it as an instance method.

Without knowing how far you got, I think you want this to be an instance method.

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/hotel.rb
```diff
@@ -1,3 +1,7 @@
 class Hotel < ApplicationRecord
    has_many :rooms
+
+   def self.guest_names
+      joins(:rooms, :guests).where().pluck(:name)
+   end
```
abdulredd, line 6: I like the idea that you had here because a join is definitely something you want. 

If we take a look at the `Hotel` model, we'll see that we have one association between a `Hotel` and a `Room`. We also know that if we take a look at the `Room` model, that a `Room` has an association with `Guest`s through `GuestRoom`.

```ruby
  has_many :rooms

  def all_guests
    guests.distinct
  end
end
```

Rails magic FTW. Play around with this and see what SQL gets printed in the terminal. A multi-join!

app/models/room.rb
```diff
@@ -1,3 +1,9 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
+
+   def guest_count
+      self.guests.distinct.count(:guest_id)
+   end
 end
```
abdulredd, line 9: 👍 

You could have also just used `count` without passing in a column name as well.

```ruby
self.guests.distinct.count
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+</ul>
+
+<h3>Add a Room:</h3>
+<div id="add_room">
+  <%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |form| %>
+    <%= form.label :room_id, "ID:" %>
+    <%= form.number_field :room_id %>
+    <%= form.submit :Submit %>
+  <% end %>
+</div>
```

```ruby
```

app/views/hotels/show.html.erb
```diff
@@ -0,0 +1,5 @@
+<h2><%= @hotel.name %></h2>
+
+<h3>All Guest Names: <% @guests.each do |guest|%>
+  <%= guest %>
+<% end %></h3>
```
abdulredd, line 5: Since we're backend devs I won't spend too much time on this, but your HTML here is incorrect.

```ruby
<h3>All Guest Names:</h3>

<% @guests.each do |guest|%>
  <p><%= guest %></p>
<% end %>
```

Beyond that, I think it would be better to, in the `Hotel` model, NOT pluck the name and just return the entire array of `Guest` objects, so that in your view you can call `<%= guest.name %>`. Just having `<%= guest %>` looks like you're printing the entire `guest` object to the view.

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,10 @@
   </head>
 
   <body>
+     <% flash.each do |type, message| %>
+      <p><%= message %></p>
+    <% end %>
+
```
abdulredd, line 16: 👍

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<h2>All Rooms:</h2>
+
+<ul id="room_info">
+<% @rooms.each do |room| %>
+    <li><strong><%= room.suite %></strong>: <%= number_to_currency(room.rate) %>, <%= room.hotel.name %>, Number of Guests: <%= room.guest_count %></li><br>
+<% end %>
+</ul>
```
abdulredd, line 7: 👍

config/routes.rb
```diff
@@ -1,3 +1,10 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/hotels/:id', to: 'hotels#show'
+  
+  get '/guests/:id', to: 'guests#show'
+
+  get '/rooms', to: 'rooms#index'
+
+  post 'guests/:id/guest_rooms', to: 'guest_rooms#create'
```
abdulredd, line 9: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221160707_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221161252_create_guest_rooms.rb
```diff
@@ -0,0 +1,8 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+    end
+  end
+end
```
abdulredd, line 8: 👍

db/seeds.rb
```diff
@@ -5,3 +5,13 @@
...
+
+  hotel_1 = Hotel.create!(name: 'The Stanley', location: 'Estes Park')
+  room_1 = hotel_1.rooms.create!(rate: 50, suite: 'Suite 1')
+  room_2 = hotel_1.rooms.create!(rate: 60, suite: 'Suite 2')
+  john = Guest.create!(name: 'John Smith', nights: 2)
+
+  GuestRoom.create!(guest: john, room: room_1)
+  GuestRoom.create!(guest: john, room: room_2)
+  
+  room_3 = hotel_1.rooms.create!(rate: 55, suite: 'Suite 3')
```
abdulredd, line 17: When you're working through and IC, it really isn't required to have to create any seed data since you will be working in the test files for the most part. If you wanted to see what the page looked like in the browser, you would just use `save_and_open_page`

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect(current_path).to eq("/guests/#{john.id}")
+        expect(page).to have_content("Room created!")
+
+        within 'ul#room_info' do
+          expect(page).to have_content("#{room_3.suite}: $#{room_3.rate}.00, #{hotel_1.name}")
+        end
+      end
+    end
+  end
+end
```
abdulredd, line 56: 👍 

Great job using 'within' in ALL of your tests! This is awesome. You also added interpolation so that you made sure the text showed up on the page exactly like you expected it to and there is no ambiguity.

I noticed that even though you included a sad path branch in your `GuestRoomsController` you didn't actually create a test for it.

---

```ruby
expect(page).to have_button(:Submit")
```

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    describe 'when I visit the rooms index page' do
+      it 'displays all of the unique guests that have stayed at this hotel' do
+        expect(page).to have_content("The Brown Palace")
+        
+        expect(page).to have_content("All Guest Names: #{john.name}, #{jane.name}")
+        expect(page).to have_content("All Guest Names: #{john.name}, #{jane.name}, #{james.name}")
+      end
+    end
+  end
+end
```
abdulredd, line 35: 👍

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        within 'ul#room_info' do
+          expect(page).to have_content("#{room_1.suite}: $#{room_1.rate}.00, #{hotel_1.name}, Number of Guests: 2")
+          expect(page).to have_content("#{room_2.suite}: $#{room_2.rate}.00, #{hotel_2.name}, Number of Guests: 1")
+          expect(page).to have_content("#{room_3.suite}: $#{room_3.rate}.00, #{hotel_2.name}, Number of Guests: 1")
+        end
+      end
+    end
+  end
+end
```
abdulredd, line 36: 👍

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  describe 'relationships' do
+    it {should belong_to :guest}
+    it {should belong_to :room}
+  end
+end
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it {should have_many :guest_rooms}
+    it {should have_many(:rooms).through(:guest_rooms)}
+  end
+end
```

spec/models/room_spec.rb
```diff
@@ -3,5 +3,38 @@
...
+      expect(room_3.guest_count).to eq(1)
+
+      james = Guest.create!(name: 'James Smith', nights: 2)
+      GuestRoom.create!(guest: james, room: room_2)
+
+      expect(room_2.guest_count).to eq(2)
+      expect(room_2.guest_count).to_not eq(1)
+    end
   end
 end
```
abdulredd, line 40: 👍

PR #20, opened 2023-02-21
Huy Phan

HuyPhan2025:
It was very hard for me and I need to get better at the set up and also have a better understanding of where I am at.  I only got half way to user story 2.

### DIFF
--- a/app/controllers/guests_controller.rb
+++ b/app/controllers/guests_controller.rb
@@ -1,0 +1,7 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
--- a/app/controllers/hotels_controller.rb
+++ b/app/controllers/hotels_controller.rb
@@ -1,0 +1,3 @@
+class HotelsController < ApplicationController
+  
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :room_guests
+  has_many :rooms, through: :room_guests
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :room_guests
+   has_many :guests, through: :room_guests
 end
--- a/app/models/room_guest.rb
+++ b/app/models/room_guest.rb
@@ -1,0 +1,4 @@
+class RoomGuest < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,17 @@
+All Guests
+
+name: <%= @guest.name %>
+
+<p> Rooms </p>
+<ul>
+  <% @rooms.each do |room| %>
+    <li>suite: <%= room.suite%>, rate: <%= room.rate %> Hotel Name: <%= room.hotel.name%>
+  <% end %>
+</ul>
+
+Add Room:
+<% form_with url: "guests/#{@guest.id}", method: :post, local: true do |f| %>
+  <%= f.label "ID: "%>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit"%>
+  <% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,3 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/guests/:id', to: 'guests#show'
 end
--- a/db/migrate/20230221160837_create_guests.rb
+++ b/db/migrate/20230221160837_create_guests.rb
@@ -1,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      
+      t.timestamps
+    end
+  end
+end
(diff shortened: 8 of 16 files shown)


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```
abdulredd, line 7: 👍

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,3 @@
+class HotelsController < ApplicationController
+  
+end
```
abdulredd, line 3: Looks like you were on your way to try and tackle the extension before you had the 3 user stories complete. I would hold off on the extension until all your user stories are complete.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :room_guests
+  has_many :rooms, through: :room_guests
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :room_guests
+   has_many :guests, through: :room_guests
 end
```

app/models/room_guest.rb
```diff
@@ -0,0 +1,4 @@
+class RoomGuest < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <li>suite: <%= room.suite%>, rate: <%= room.rate %> Hotel Name: <%= room.hotel.name%>
+  <% end %>
+</ul>
+
+Add Room:
+<% form_with url: "guests/#{@guest.id}", method: :post, local: true do |f| %>
+  <%= f.label "ID: "%>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit"%>
+  <% end %>
```
abdulredd, line 17: We mustn't forget about the HTML. Always put rendered ERB in an HTML tag.

```ruby
<h1>All Guests</h1>

<p>name: <%= @guest.name %></p>

<p>Add Room:</p>
```

---

The route in your form is not a RESTful route. Currently, the way that the route is formatted it looks something like this:

```ruby
post "/guests/:id"
```

This does not follow RESTful convention as the path for a POST never ends in an ID. The current route have is for `show`, `update`, and `delete` actions.

We need to remember what action we are trying to perform using the form. We want to create an association between a room and a guest. So essentially create a new entry in the RoomGuests join table. Your route should look something like this:

```ruby
<%= form_with url: "guests/#{@guest.id}/rooms", method: :post, local: true do |f| %>
```

This route would resolve to something like this: `post "guests/:id/rooms"`. One more thing, your form needs to be a rendered ERB tag with the `=`, right now as written, your form wouldn't show up on the page.

---

Last thing, the `f.label` should also have a name attribute that is the same as the field that the label is being used for.

```ruby
```

config/routes.rb
```diff
@@ -1,3 +1,3 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-end
+  get '/guests/:id', to: 'guests#show'
+end
```
abdulredd, line 3: 👍 

I'm assuming that this is where you ran out of time and were unable to add the route for the form. But as I mentioned in the comment in that view, you want the route to be `post "/guests/:id/rooms", to: "room_guests#create"` and not be going to `post "/guests/:id", to: "guests#create"` (I'm assuming this route based on the route you had in your form)

db/migrate/20230221160837_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221170705_create_room_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateRoomGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :room_guests do |t|
+        t.references :room, foreign_key: true
+        t.references :guest, foreign_key: true
+        
+        t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it "when I fill in the form I should be redirect back to the guest show page with the added room" do
+      fill_in :room_id, with: "#{@super.id}"
+      click_button "Submit"
+
+      expect(current_path).to eq("/guests/#{max.id}")
+      expect(page).to have_content("Super")
+    end
+  end
+
+end
```
abdulredd, line 45: 👍 

```ruby
```

---

One other thing, on line `37` you are referencing a variable called `@super` which doesn't exist since you are not using a `before :each` and instead using `let!`...

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it {should have_many :room_guests}
+    it {should have_many(:rooms).through(:room_guests)}
+  end
```
abdulredd, line 7: 👍

spec/models/room_guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe RoomGuest, type: :model do
+  describe 'relationships' do
+    it { should belong_to :guest }
+    it { should belong_to :room }
+  end
+end
```
abdulredd, line 8: 👍

spec/models/room_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 RSpec.describe Room, type: :model do
   describe 'relationships' do
-    it {should belong_to :hotel}
+    it { should belong_to :hotel }
+    it { should have_many :room_guests }
+    it { should have_many(:guests).through(:room_guests) }
   end
 end
```

PR #21, opened 2023-02-21
Andra Helton

ALHelton:
Got through story 3

I feel that I needed a bit more time with this challenge - I intended to create a more robust test for my instance method guest_count but ran out of time.


app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def index
+    @rooms = Room.all
+  end
+
+  def create
+    guest_room = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id], )
+
+    if !guest_room.save
+      flash[:alert] = et.errors.full_messages
```
cjsim89, line 11: Nice job writing a flash message here, but just so you know you were not required to implement any sad path for this challenge, and won't be for the final either. But this is good practice.

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,16 @@
+class RoomsController < ApplicationController
+
+  def index
+    @rooms = Room.all
+  end
+
+  def create
+    guest_room = GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id], )
```
cjsim89, line 8: I'm not sure I would put this `GuestRoom.create` in the `RoomsController`, since what we're creating is really a `GuestRoom` and not a `Room` on its own. For this, I think it would be better to make a `GuestRoomsController` file separately, and use the `create` method there instead. But the actual "meat" of this method, its functionality, is perfect.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <br>
+  Suite: <%= room.suite %>
+  <br>
+  Rate: <%= room.rate %>
+  <br>
+  <br>
+<% end %>
+
+<h2>Add a Room to this Guest:</h2>
+<%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |f| %>
```
cjsim89, line 15: Great job with this form! Looking RESTful here.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe 'Show', type: :feature do
+  
+  let!(:echo) { Hotel.create!(name: 'Echo Mountain Inn', location: 'Echo Mountain') }
+  let!(:winterpark) { Hotel.create!(name: 'Winter Park Hotel', location: 'Winter Park') }
+
+  let!(:andra) { Guest.create!(name: "Andra", nights: 4) }
```
cjsim89, line 8: Nice job with this data setup for your tests. I would try to make at least 1 more Guest that you can use to play with in your scenarios; one is ok but more data is always good 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+            expect(page).to have_content("Add a Room to this Guest:")
+
+            expect(page).to have_field(:room_id)
+            expect(page).to have_button("Submit")
+
+            fill_in :room_id, with: e42.id
+            click_on "Submit"
+
+            expect(current_path).to eq("/guests/#{andra.id}")
+            expect(page).to have_content("Hotel: Winter Park Hotel")
```
cjsim89, line 54: This test is ok, one thing to make it stronger would be to test the _negative_ case of this test; this line is really the important part since this is the Room that we're adding to this guest. So, if at the top of this `it` block we added 
```
expect(page).to_not have_content("Winter Park Hotel")
```
and it passes, then we know it's working when it's not there before but there after.

spec/models/room_spec.rb
```diff
@@ -1,7 +1,28 @@
...
+
   describe 'relationships' do
     it {should belong_to :hotel}
+    it { should have_many(:guests).through(:guest_rooms) }
+  end
+
+  describe 'Instance Methods' do
+    describe '#guest_count' do
+      it 'totals the number of guests' do
+        expect(b27.guest_count).to eq(2)
```
cjsim89, line 24: I would like to see one expect line per Room you create... and there should be more than 1 room 😄 I agree that this model test could be beefed up a bit,.

PR #22, opened 2023-02-21
Sam Walker

sgwalker327:
Sam Walker
I feel like it went well. I went a couple minutes over sorting out a syntax error.
User Story 3


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,7 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guestroom = GuestRoom.new(room_id: params[:room_id], guest_id: params[:id])
+    guestroom.save
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```

The only thing I would probably do here, since you are not doing any flash messaging, would be to probably use `create` over `new`+`save`.

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+end
```

app/controllers/hotel_controller.rb
```diff
@@ -0,0 +1,3 @@
+class HotelsController < ApplicationController
+
+end
```
abdulredd, line 3: Looks like you were on your way to try and tackle the extension. Don't forget to use TDD though. I don't see a `features/hotels/show_spec`.

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,5 @@
+class RoomsController < ApplicationController
+  def index
+    @rooms = Room.all
+  end
+end
```

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,10 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
+
+   def guest_count
+    # require 'pry'; binding.pry
+    guests.distinct.count
+   end
```
abdulredd, line 9: 👍 

I think the logic for the `guest_count` method could have been implemented with or without `distinct` according to each person's understanding of how guests should be counted.

For example, if there's a room and I stay in it, then next week you stay in the same room, that's obviously 2 guests that have stayed in that room. But if I come back and stay in that room a second time, it is still 2 guests or 3 guests and every stay is counted as its own thing 🤷 .

Long way to say, I don't think the user story was specific about this, and as long as there was no ruby being used, both cases would be accepted. Great job!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+Nightly Rate: <%= number_to_currency(room.rate) %>
+<br>
+<% end %>
+
+Add Room:
+<%= form_with url: "/guests/#{@guest.id}/guest_rooms", method: :post, local: true do |f| %>
+  <%= f.label "Room ID:" %>
+  <%= f.number_field :room_id %>
+  <%= f.submit "Submit" %>
+<% end %>
```
abdulredd, line 17: We mustn't forget about the HTML. Always put rendered ERB in an HTML tag.

```ruby
<h1><%= @guest.name %></h1>

<p>Room Stayed Im:</p>

<p><%= room.hotel.name %> - <%= room.suite %></p>

<p>Nightly Rate: <%= number_to_currency(room.rate) %></p>

<p>Add Room:</p>
```

---

I already mentioned this in the `routes.rb` file, but the URL for the form isn't correct according to REST.

```ruby
```

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Room Index:</h1>
+
+<% @rooms.each do |room| %>
+<div id= "room-<%= room.id %>" >
+<%= room.suite %> - <%= room.hotel.name %>
+Nightly Rate: <%= number_to_currency(room.rate) %>
+Number of guests: <%= room.guest_count %>
+</div>
+<% end %>
```
abdulredd, line 9: Same comment I made before, this is an HTML doc, only non-rendered ERB shouldn't be surrounded by HTML.

config/routes.rb
```diff
@@ -1,3 +1,7 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/guests/:id', to: 'guests#show'
+  post '/guests/:id/guest_rooms', to: 'guest_rooms#create'
+
+  get '/rooms', to: 'rooms#index'
 end
```
abdulredd, line 7: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221160932_create_guests.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuests < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/migrate/20230221162428_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        click_on "Submit"
+
+        expect(current_path).to eq("/guests/#{steve.id}")
+        expect(page).to have_content("High Roller")
+        expect(page).to have_content("Nightly Rate: $200.00")
+        expect(page).to have_content("Tyrolean Inn")
+      end
+    end
+  end
+end
```
abdulredd, line 52: 👍 

```ruby
```

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+          expect(page).not_to have_content("Number of guests: 2")
+        end
+
+        within "#room-#{shack.id}" do
+          expect(page).to have_content("Number of guests: 2")
+          expect(page).not_to have_content("Number of guests: 1")
+        end
+      end
+  end
+end
```
abdulredd, line 44: 👍

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it {should have_many :guest_rooms}
+    it {should have_many(:rooms).through(:guest_rooms)}
+  end
+end
```

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom do
+  describe 'relationships' do
+    it {should belong_to :guest}
+    it {should belong_to :room}
+  end
+end
```

spec/models/room_spec.rb
```diff
@@ -1,7 +1,27 @@
...
+    it {should have_many(:guests).through(:guest_rooms)}
+  end
+
+  describe 'Instance Methods' do
+    it '#guest_count' do
+      expect(bum.guest_count).to eq(2)
+      expect(bum.guest_count).not_to eq(3)
+    end
   end
 end
```
abdulredd, line 27: 👍 

Great choice making this an instance method on the `Room` model. I think that is the easiest approach for what is being asked. The only thing I would say to add is another assertion to verify that the value, `2`, isn't hard-coded in the `guest_count` method.

Remember that when we are following TDD, it says that we should write the "smallest amount of code in order to get the test to pass"... in this case, the smallest amount of code would be to just hard-code the value `2` in the method.

```ruby
def guest_count
  2
end
```

Just by looking at the test, there's no way to tell is the logic is hard-coded or if the method is actually performing some logic. In order to make it obvious I like to add another assertion with an updated value.

```ruby
expect(bum.guest_count).to eq(2)
expect(bum.guest_count).not_to eq(3)

let!(:sam) { Guest.create!(name: "Sam", nights: 2) }
GuestRoom.create!(guest: sam, room: bum)

expect(bum.guest_count).to eq(3)
```

With this change, we've seen the value of `guest_count` change from 2 to 3, expressing that the logic is not hard-coded and updates as the data changes.

PR #23, opened 2023-02-21
Matt Enyeart BE 2211

menyeart:
I think I got through story 4.
One issue I wasn't sure about was how to test the guest model method that returned only unique values(guests). For mine I wanted to return objects so I couldn't figure out how to test for duplicates because the database can't have a duplicate objects via id. I tried adding them to an array and calling the method on the array, but it didn't seem to work.


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,9 @@
+class GuestRoomsController < ApplicationController
+  def create
+    guest = Guest.find(params[:id])
+    room = Room.find(params[:room_id])
+    GuestRoom.create!(guest: guest, room: room)
```
cjsim89, line 5: Nice job creating this GuestRooms controller, it's exactly the approach I like to take when creating these join objects. 🙌

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,6 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @unique_guests = @hotel.guests.uniq_guests
```
cjsim89, line 4: This is fine, but you can also call `.uniq_guests` on the `@hotel.guests` association in the view. Either way is fine.

app/models/hotel.rb
```diff
@@ -1,3 +1,4 @@
 class Hotel < ApplicationRecord
    has_many :rooms
+   has_many :guests, through: :rooms
```
cjsim89, line 3: Nice job making this association! Will make your joining a lot easier.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<h1><%= @guest.name %></h1>
+<h3> Rooms: </h3>
+<% @guest.rooms.each do |room| %>
+  <p><%= "Suite: #{room.suite}, Nightly Rate: #{room.rate}, Hotel: #{room.hotel.name} " %></p>
+<% end %>
+
+<h3>Add an existing room (by id) to this guest: </h3>
+  <%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |form| %>
```
cjsim89, line 8: Nice form setup here! I think this kind of path is good, the alternative you can do with this kind of operation is 
```
url: "/guest_rooms"
```
and have the `guest_id` inside the form as a hidden field.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      it "it has a form to add a room to this guest. When I fill in the fields and click submit I am redirected back to this guest's show page and I see the room now listed under this guest" do
+        stanley = Hotel.create!(name: "The Stanley", location: "Estes Park")
+        grand_budapest = Hotel.create!(name: "The Grand Budapest Hotel", location: "Budapest")
+        pres = Room.create!(suite: 'Presidential', rate: 400, hotel: stanley)
+        king = Room.create!(suite: 'King', rate: 500, hotel: stanley)
+        poor = Room.create!(suite: 'Poor', rate: 20, hotel: grand_budapest)
+        matt = Guest.create!(name: 'Matt', nights: 3)
+        guestroom1 = GuestRoom.create!(guest: matt, room: pres )
+        guestroom2 = GuestRoom.create!(guest: matt, room: king )
```
cjsim89, line 32: This test setup data looks good; you may want to consider using a `before :each ` block under your 2nd `describe` block, just to save some time. Totally up to you though; I tend to use both/either, wherever it fits best.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        visit "/guests/#{matt.id}"
+
+        expect(page).to_not have_content("The Grand Budapest Hotel")
+        expect(page).to_not have_content("Poor")
+        expect(page).to_not have_content("20")
```
cjsim89, line 38: Really good test setup here, looking for the "negative" case before the "positive" to add weight. Nice job!

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "can count the total number of guests" do
+      matt = Guest.create!(name: 'Matt', nights: 5)
+      keith = Guest.create!(name: 'Keith', nights: 2)
+      stephanie = Guest.create!(name: 'Stephanie', nights: 4)
+
+      expect(Guest.count_guests).to eq(3)
+
+      sunita = Guest.create!(name: 'Sunita', nights: 4)
+
+      expect(Guest.count_guests).to eq(4)
```
cjsim89, line 19: So, this is a tricky method to test... You implemented `count_guests` as a Class method on `Guest`. This is fine, but when we do this we have 2 different ways we can call the method --- on a collection of ALL the guests, or a collection of SOME of the guests based on an associated Room. The 2nd way is what the user story asks for, and it's how you have it implemented in your View and Feature tests, which is great. 
However, we also need to be sure we're testing for that in the model test - the model tests are unit tests that ensure the functionality we deisgned is working the way it's intended. This test is looking at calling on ALL the guests, we just need to have additional tests to call on specific associations of Rooms. 

So, if this included 
```
room1 = Room.create(...)
room2 = Room.create(...)

GuestRoom.create(guest: matt, room: room1)
GuestRoom.create(guest: keith, room: room1)

```
then that would be complete. As it is though, I would say this model is not completely tested.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "return a list of unique guests" do
+      matt = Guest.create!(name: 'Matt', nights: 5)
+      keith = Guest.create!(name: 'Keith', nights: 2)
+      stephanie = Guest.create!(name: 'Stephanie', nights: 4)
+
+      expect(Guest.uniq_guests).to eq([matt, keith, stephanie])
+
+      sunita = Guest.create!(name: 'Sunita', nights: 4)
+
+      expect(Guest.uniq_guests).to eq([matt, keith, stephanie, sunita])
```
cjsim89, line 31: Similarly to above, the extension asks for this list to be all the guests that stayed _at this Hotel_. I think having an instance of Hotel that we've added some Rooms to and then some Guests to those Rooms, would make this model test complete for the way you designed its functionality. 

Alternatively, you could have made this an instance method on Hotel - and since you gave it another association of `has_many :guests, through: :rooms`, it makes the method very simple to write: 
```
guests.distinct
```
And if you make it this instance method, it makes the testing a lot easier as well. 

So, be careful that your model tests reflect the way you have actually _implemented_ the methods, and not just the way it _can_ work.

cjsim89: Also, from your question in the reflection, this test doesn't quite test for "uniqueness"... Consider what the user story is asking for. A hotel should be able to return a unique list of its guests. That means, we'd need to set up a scenario where a guest has stayed at this hotel _more than once_; and in that scenario,t he list would only return that name once. 
This is another argument for having this method be an instance method of Hotel (see my comment above).

cjsim89 (commented):
(Added another comment to the `unique_guest` test)

PR #24, opened 2023-02-21
David Marino

davejm8:
User Story 3 brought me down a rabbit hole. I was looking for a way for this to populate a number that wouldn't drop once a guest leaves the hotel. I ended up going back to simplify my answer but ran out of time.


app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,8 @@
+class GuestRoomsController < ApplicationController
+
+  def create
+    GuestRoom.create!(guest_id: params[:id], room_id: params[:room_id])
+
+    redirect_to "/guests/#{params[:id]}"
+  end
+end
```
abdulredd, line 8: 👍

app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestsController < ApplicationController
+
+  def show
+    @guest = Guest.find(params[:id])
+  end
+end
```

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class RoomsController < ApplicationController
+  
+  def index
+    @rooms = Room.all
+  end
+end
```
abdulredd, line 6: 👍

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

db/migrate/20230209221054_create_rooms.rb
```diff
@@ -3,6 +3,7 @@ def change
     create_table :rooms do |t|
       t.integer :rate
       t.string :suite
+      t.integer :reservations
```
abdulredd, line 6: I wonder why you felt the need to add a new column to the rooms table. You shouldn't have had to do this, since we can get the number of guests in a room through the GuestRooms join table.

🤷

app/models/guest_room.rb
```diff
@@ -0,0 +1,10 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+  after_create :reservation_count
+
+  def reservation_count
+    self.room.reservations = self.room.reservations.to_i + 1
+    self.save
+  end
```
abdulredd, line 9: I think we are maybe overcomplicating what the user story is asking from us. All the user story wants to know is how many people have stayed in the rooms and we can easily do that through the GuestRooms joins table since the  Room model has an association to Guests through the GuestRooms join table. No new `reservations` column needed or `after_create` callback.

We're just adding extra overhead to something that we already have implementation on.

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
 end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= @guest.name %>
```
abdulredd, line 1: Don't forget to use the HTML knowledge you gained through the intermission work.

```ruby
<h1><%= @guest.name %></h1>
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<p> Rooms Stayed In:</p>
+<ul>
+<% @guest.rooms.each do |room| %>
+<li> <%= room.suite %> <%= room.id %> ($<%= room.rate %> at the <%= room.hotel.name %>) </li>
+<% end %>
+</ul>
+</div>
+
+<h2>Add another room</h2>
+<%= form_with url: "/guest_rooms/#{@guest.id}", method: :post, local: true do |f| %>
```
abdulredd, line 11: This is not a RESTful route. Currently the way that the route is formatted it looks something like this:

```ruby
post "/guest_rooms/:id"
```

This does not follow RESTful convention as the path for a POST never ends in an ID.

If you did someting like: `url: "/guest_rooms"` then that would be ok, but you would need to add a hidden field to the form where you put the guests ID.

```ruby
<%= f.hidden_field :guest_id, value: @guest.id %>
```

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,7 @@
+
+<% @rooms.each do |room| %>
+  <h1><%= room.suite %></h1>
+  <p>Rate: <%= room.rate %></p>
+  <p>Hotel: <%= room.hotel.name %>
+  <p>Rented: <%= room.guests.count %></p>
+<% end %>
```
abdulredd, line 7: 👍

config/routes.rb
```diff
@@ -1,3 +1,9 @@
 Rails.application.routes.draw do
+
+  get '/rooms', to: 'rooms#index'
+
+  get '/guests/:id', to: 'guests#show'
+
+  post '/guest_rooms/:id', to: 'guest_rooms#create'
```
abdulredd, line 7: Like I mentioned in the comment in the view, this is not a RESTful route. Remember that only show, delete, update have paths that end with an ID, create routes do not. The 2 options that you could have used here were:

```ruby
post "/guest_rooms", to: "guest_rooms#create"
```

or

```ruby
post "guests/:id/rooms", to: "guest_rooms#create"
```

The second one is more common, especially when we're talking about using the `resources` method and nesting resources. Something like this would produce the second route:

```ruby
resources :rooms, only: :index

  resources :rooms, only: :create, controller: "guest_rooms"
end
```

db/migrate/20230221161247_create_guests.rb
```diff
@@ -0,0 +1,11 @@
...
+  def change
+    create_table :guests do |t|
+      t.string :name
+      t.integer :nights
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 11: 👍

db/migrate/20230221161418_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: 👍

db/seeds.rb
```diff
@@ -5,3 +5,17 @@
...
+room_3 = holliday.rooms.create!(rate: 160, suite: "Honeymoon")
+room_4 = holliday.rooms.create!(rate: 10, suite: "Broom Closet")
+steve = Guest.create!(name: 'Steve', nights: 3)
+dave = Guest.create!(name: 'Dave', nights: 5) 
+jim = Guest.create!(name: 'Jim', nights: 1)
+GuestRoom.create!(guest: steve, room: room_1)
+GuestRoom.create!(guest: dave, room: room_1)
+GuestRoom.create!(guest: steve, room: room_2)
+GuestRoom.create!(guest: dave, room: room_3)
+GuestRoom.create!(guest: dave, room: room_4)
```
abdulredd, line 21: When you're working through and IC, it really isn't required to have to create any seed data since you will be working in the test files for the most part. If you wanted to see what the page looked like in the browser, you would just use `save_and_open_page`

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    expect(page).to have_content(room_2.rate)
+    expect(page).to have_content(harbor.name)
+
+    visit "/guests/#{dave.id}"
+
+    expect(page).to have_content(dave.name)
+    expect(page).to have_content(room_3.suite)
+    expect(page).to have_content(room_3.rate)
+    expect(page).to have_content(holliday.name)
+  end
```
abdulredd, line 38: I believe, in order to make your rests a little more robust, to do some interpolation on what you are expecting on the page. Example:

```ruby
expect(page).to have_content("#{room_1.suite} #{room_1.id} ($#{ room.rate} at the #{room.hotel.name})")
```

This way we are ensured that we are looking at the right content and that `room_1.rate` that we are expecting to be on the page isn't just some random number that RSpec so on the page as maybe part of a date or something like that and marked as being found.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    expect(page).to have_content(room_3.suite)
+    expect(page).to have_content(room_3.rate)
+
+    fill_in 'Find room with ID', with: room_4.id
+    click_button 'Add Room'
+
+    expect(page).to have_content(room_4.suite)
+    expect(page).to have_content(room_4.rate)
+  end
```
abdulredd, line 55: The only thing missing in this test is the check for the redirect according to the user story:

```ruby
expect(current_path).to eq("/guests/#{steve.id}")
```

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    
+    expect(page).to have_content(room_2.suite)
+    expect(page).to have_content(room_2.rate)
+    expect(page).to have_content(room_2.guests.count)
+
+    expect(page).to have_content(room_3.suite)
+    expect(page).to have_content(room_3.rate)
+    expect(page).to have_content(holliday.name)
+    expect(page).to have_content(room_3.guests.count)
+  end
```
abdulredd, line 40: Like I mentioned in your previous test, interpolation tests are better and let's say you couldn't properly do some interpolation in the assertions. You could then use some `within` blocks. The most important thing is that we want to be absolutely sure that the things were are asserting to be true are actually true and not false-positive.

```ruby
within "#room-#{@room_1.id}" do
  expect(page).to have_content(@room1.suite)
  expect(page).to have_content(@room1.rate)
  expect(page).to have_content(@hotel1.name)
end
```

This would mean that you would need to update your view to something like this:

```ruby
<% @rooms.each do |room| %>
  <div id="room-<%= room.id %>">
    <h1><%= room.suite %></h1>
    <p>Rate: <%= room.rate %></p>
    <p>Hotel: <%= room.hotel.name %>
    <p>Rented: <%= room.guests.count %></p>
  </div>
<% end %>
```

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    GuestRoom.create!(guest: steve, room: room_2)
+
+    GuestRoom.create!(guest: dave, room: room_3)
+
+    GuestRoom.create!(guest: dave, room: room_4)
+
+    expect(page).to have_content(room_1.suite)
+    expect(page).to have_content(room_1.rate)
+    expect(page).to have_content(harbor.name)
+    expect(page).to have_content(room_1.guests.count)
```
abdulredd, line 29: This test is failing. Whenever the test reaches line `29` it fails because it does not find the number `3` on the page, `3` being the number of guests who have rented room 1.

The issue here is ordering, because we `visit` the page before the `GuestRoom` associations get created, there is only 1 guest who has stayed in the room at the time of the request. The solution is to move the `visit` to after the `GuestRoom` associations.

```ruby
GuestRoom.create!(guest: steve, room: room_1)
GuestRoom.create!(guest: dave, room: room_1)

GuestRoom.create!(guest: steve, room: room_2)

GuestRoom.create!(guest: dave, room: room_3)

GuestRoom.create!(guest: dave, room: room_4)

visit '/rooms'
```

By this time, when the request is made, we would have those associations in the database.

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    GuestRoom.create!(guest: dave, room: room_3)
+    GuestRoom.create!(guest: dave, room: room_4)
+  end
+
+      it 'counts the number of guests who have rented the room' do
+        expect(room_1.reservations).to eq(2)
+      end
+    end
+  end
+end
```
abdulredd, line 32: I think we are overcomplicating things here a bit. I think what we want is an instance method on the `Room` model that counts all the guests through the association it has with the GuestRooms table.

On your `Room` model:

```ruby
def guest_count
  guests.count
end
```

And that would have been all you needed. There may be good reasoning behind adding a method to the join model, but I don't think this was one of them.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  it {should have_many :guest_rooms}
+  it {should have_many(:rooms).through(:guest_rooms)}
+end
```
abdulredd, line 6: 👍

spec/models/room_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Room, type: :model do
   describe 'relationships' do
     it {should belong_to :hotel}
+    it {should have_many :guest_rooms}
+    it {should have_many(:guests).through(:guest_rooms)}
   end
 end
```
abdulredd, line 9: This test file would be where I would expect a `#guest_count` instance method test.

PR #25, opened 2023-02-21
Isaac Alter Mid Mod 2 Assessment

Isaac3924:
Was moving slower than I liked, but I found I had a good grasp on the content needed to pass the assessment's expectations. Unfortunately, I found my knowledge on joins was lacking once I reached extension 1. I would like a more solid basis on that, but otherwise, I feel I followed restful and mvc guidelines as was taught.

I made it through Story 3. I have commented out code relating to extensions 1.


app/controllers/guests_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  def show
+    @guest = Guest.find(params[:id])
+    @rooms = @guest.rooms
+  end
+
+  def update
+    @guest = Guest.find(params[:id])
+    @room = Room.find(params[:search])
+    GuestRoom.create!(guest: @guest, room: @room)
+    redirect_to "/guests/#{@guest.id}"
```
cjsim89, line 11: While the functionality of this method is correct, I disagree that we're updating a Guest record in this method - the "meat" of the functionality is the `GuestRoom.create!` where we're actually creating a new _relationship_ between the existing Guest and Room records, by their respective IDs. 

Because of this, I recommend putting this in a `GuestRoomsController` instead, using the `create` action there since that's what we're doing with a GuestRoom.

app/controllers/hotels_controller.rb
```diff
@@ -0,0 +1,7 @@
+class HotelsController < ApplicationController
+  def show
+    @hotel = Hotel.find(params[:id])
+    @rooms = @hotel.rooms
```
cjsim89, line 4: This is fine, but you're able to call `hotel.rooms` in a view. This would also keep your controller "skinny" and not duplicate data that's already available.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1><%= @guest.name %></h1>
+<ul><% @rooms.each do |room| %>
+  <li><%= room.suite %>- Rate: <%= room.rate %>, Hotel: <%= room.hotel.name %></li>
+<% end %>
+</ul>
+
+<%= form_tag("/guests/#{@guest.id}", method: :patch, local: true) do %>
```
cjsim89, line 7: Using `form_tag` is an outdated way to create a form using a ViewHelper. We recommend using `form_with`, and we'll be adding to the features of `form_with` in class this week, in a way that is not accessible in a `form_tag` method.

cjsim89: Having said that, your form construction is good - overall try to use `form_with` and corresponding helper methods inside of it from now on.

config/routes.rb
```diff
@@ -1,3 +1,8 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :hotels, only: [:show]
+
+  resources :rooms, only: [:index]
+
+  resources :guests, only: [:show]
```
cjsim89, line 6: Nice job using `resources` syntax!

config/routes.rb
```diff
@@ -1,3 +1,8 @@
...
+  patch '/guests/:id', to: 'guests#update'
```
cjsim89, line 7: As I said above, I would rather you use `:post` to create a new relationship between a Guest and a Room. 
That route might look like: 
```
post '/guest_rooms', to: 'guest_rooms#create'
```

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect(page).to have_content(@room1.rate)
+    expect(page).to have_content(@room2.rate)
+    expect(page).to have_content(@room3.rate)
+
+    expect(page).to have_content("Holiday Inn")
+    expect(page).to have_content("Mariott")
+    expect(page).to have_content(@hotel1.name)
+    expect(page).to have_content(@hotel2.name)
+
+    expect(page).to_not have_content("575")
```
cjsim89, line 60: These tests overall look good; try to be more specific with values like this; what is "575" in this context? If in your view you have "Rate: 575", then you should be testing that in your feature test here.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect(page).to_not have_content(@room4.rate)
+    expect(page).to_not have_content("Condor")
+    expect(page).to_not have_content(@hotel3.name)
+  end
+
+  scenario "I see a form to add a room to this guest" do
+    fill_in 'search', with: "#{@room4.id}"
```
cjsim89, line 67: It might be good to first test that the form field is there, by doing something like: 
```
expect(page).to have_field "search"
```

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  scenario "I fill in the field with an id of an existing room and click submit and I see the room now listed" do
+    expect(page).to_not have_content("575")
+    expect(page).to_not have_content(@room4.rate)
+    expect(page).to_not have_content("Condor")
+    expect(page).to_not have_content(@hotel3.name)
+
+    fill_in 'search', with: "#{@room4.id}"
+    click_button 'Submit'
+
+    expect(page).to have_content("575")
```
cjsim89, line 88: Good job using `.to_not` first before `.to`, so that you add weight to the positive test cases at the end of this scenario.

spec/features/hotels/show_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+  
+  # scenario "I see a unique list of guest names" do
+  #   expect(page).to have_content("Bob").once
+  #   expect(page).to have_content(@guest1.name).once
+  #   expect(page).to_not have_content("Shane").once
+  #   expect(page).to_not have_content(@guest2.name).once
+  #   expect(page).to_not have_content("Isaac").once
+  #   expect(page).to_not have_content(@guest3.name).once
+  # end
```
cjsim89, line 37: This looks like a good start to the extension, try to finish this out for good practice!

app/models/guest.rb
```diff
@@ -0,0 +1,8 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+
+  def self.count_guests
+    count(:id)
```
cjsim89, line 6: This method works, but you will want ot be careful how this method is both tested & implemented, since hte user story is asking you to find `the number of guests that have stayed in that room.`. Since you created a class method, you will only want to call this on a Room instance's Guests collection. 

This would be a little easier to test and implement if it were an instance method on Room, that used the `guests` association.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1><%= @guest.name %></h1>
+<ul><% @rooms.each do |room| %>
+  <li><%= room.suite %>- Rate: <%= room.rate %>, Hotel: <%= room.hotel.name %></li>
```
cjsim89, line 3: This doesn't necessarily match what you have in your test, where the test just says the values not related to the keys ("Rate: ", "Hotel:" ). Be careful that your feature tests echo what the view is exactly.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  describe "class methods" do
+    it "#count" do
+      guest1 = Guest.create!(name: "Bob", nights: 3)
+      guest2 = Guest.create!(name: "Shane", nights: 10)
+
+      expect(Guest.count_guests).to eq(2)
+
+      guest3 = Guest.create!(name: "Isaac", nights: 13)
+
+      expect(Guest.count_guests).to eq(3)
```
cjsim89, line 18: The danger of using a class method in this scenario is that there are now 2 ways it can be used - called on a collection of ALL the guests, or called on a collection of the guests _associated to a Room_. The 2nd way is what the user story is asking for, so we should really be testing it in that way as well. As is, this doesn't fully cover the functionality of this method. 

What would make it fully covered would be to add some `Room` instances and relate some Guests to those Rooms already, and make sure the value returned is different for each Room instance. Example: 
```
room1 = Room.create!(...)
room2 = Room.create!(...)

GuestRoom.create(room: room1, guest: guest1)
GuestRoom.create(room: room1, guest: guest2)
GuestRoom.create(room: room2, guest: guest1)
```
Then our expect statements would include:
```
expect(room2.guests.count_guests).to eq(1)
```

cjsim89: Doing it this way ^ like I illustrated would cover the use case that the user story is asking for, and that's what I want to underline is important about writing model tests. You did cover this functionality in the feature test, but that's for intergrations - model tests are meant to be unit/bite-size/atomic tests that ensure functionality works in the way we intend it. Since you used it on an association and not on a Model in the application itself, we need to ensure that way is tested. Let me know if you want to talk about this more.

PR #26, opened 2023-02-21
James .T | Mid-Mod

JTaylor28:
I felt like I had this, but I had too many migration errors in the beginning and lost a lot of time due to that. For whatever reason my rails db:migrate created tables from my practice IC  that i didn't want and it took me too long to figure out how to stop that from happening.

### DIFF
--- a/app/controllers/room_controller.rb
+++ b/app/controllers/room_controller.rb
@@ -1,0 +1,9 @@
+class RoomController < ApplicationController
+
+  def create
+    
+   guest_room = GuestRoom.new(guest_id: params[:id], room_id: params[:room_id])
+   guest_room.save
+   redirect_to "/guests/#{params[:id]}"
+  end
+end
--- a/app/models/guest.rb
+++ b/app/models/guest.rb
@@ -1,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
--- a/app/models/room.rb
+++ b/app/models/room.rb
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,7 +1,14 @@
 <h1><%= @guest.name %>'s page</h1>
 <H3>Rooms stayed in</h3>
 <% @guest.rooms.each do |room| %>
-  <p>Room:  <%= room.suite %> </p>
+  <p>Rooms:  <%= room.suite %> </p>
   <p>Rate:  <%= room.rate %> </p>
   <p>Hotel:  <%= room.hotel.name %> </p>
 <% end %>
+
+<h3>Add Room</h3>
+<%=form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |f|%>
+  <%=f.label "ID: " %>
+  <%=f.number_field :room_id %>
+  <%=f.submit "Add Room" %>
+<%end%>
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,7 @@
+<h1><%= @guest.name %>'s page</h1>
+<H3>Rooms stayed in</h3>
+<% @guest.rooms.each do |room| %>
+  <p>Room:  <%= room.suite %> </p>
+  <p>Rate:  <%= room.rate %> </p>
+  <p>Hotel:  <%= room.hotel.name %> </p>
+<% end %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  get "/guests/:id", to: "guests#show"
+  
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
 
   get "/guests/:id", to: "guests#show"
-  
+
+  post "/guests/:id/rooms", to: "room#create"
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,9 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+@hotel_1 = Hotel.create!(name: 'Stanly Hotel', location: 'Estes Park')
+@room_401 = @hotel_1.rooms.create!(rate: 50, suite: 'Room 401')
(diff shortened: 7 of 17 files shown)


config/routes.rb
```diff
@@ -1,3 +1,6 @@
 Rails.application.routes.draw do
-  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+
+  get "/guests/:id", to: "guests#show"
+
+  post "/guests/:id/rooms", to: "room#create"
```
abdulredd, line 5: Great job setting up the guest's show route correctly with the right controller and action. The post route to `"/guests/:id/rooms"` is also correct. The controller and action that you have for this URI path is what I want to focus on here.

Let's say we have this route: `post "/rooms"`, this would also go to the `RoomsController` and the `create` but we never do anything inside of the `create` action to determine what route was actually used to hit the action.

Was it `post "/guests/:id/rooms"` or `post "/rooms"`? (I know we don't have a create rooms user story right now, but we have to think about that functionality possibly being added later on)

Right now your `RoomsController` `create` action looks like this:

```ruby
def create
   guest_room = GuestRoom.new(guest_id: params[:id], room_id: params[:room_id])
   guest_room.save
   redirect_to "/guests/#{params[:id]}"
end
```

If it was a `post "/rooms"` that hit this action, this logic would fail because there would be no `:room_id`. Instead of going into how you could modify this method to work regardless of which of the path triggered the action, I would recommend you to use a join controller like `GuestRoomsController` and in there have your `create` action.

app/controllers/room_controller.rb
```diff
@@ -0,0 +1,9 @@
+class RoomController < ApplicationController
+
+  def create
+    
+   guest_room = GuestRoom.new(guest_id: params[:id], room_id: params[:room_id])
+   guest_room.save
+   redirect_to "/guests/#{params[:id]}"
+  end
```
abdulredd, line 8: I would move this logic to a joins controller like, `GuestRoomsController` instead of having it here like this.

app/models/guest.rb
```diff
@@ -0,0 +1,4 @@
+class Guest < ApplicationRecord
+  has_many :guest_rooms
+  has_many :rooms, through: :guest_rooms
+end
```

app/models/guest_room.rb
```diff
@@ -0,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :guest
+  belongs_to :room
+end
```

app/models/room.rb
```diff
@@ -1,3 +1,5 @@
 class Room < ApplicationRecord
    belongs_to :hotel
+   has_many :guest_rooms
+   has_many :guests, through: :guest_rooms
 end
```

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<H3>Rooms stayed in</h3>
+<% @guest.rooms.each do |room| %>
+  <p>Rooms:  <%= room.suite %> </p>
+  <p>Rate:  <%= room.rate %> </p>
+  <p>Hotel:  <%= room.hotel.name %> </p>
+<% end %>
+
+<h3>Add Room</h3>
+<%=form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |f|%>
+  <%=f.label "ID: " %>
```
abdulredd, line 11: Don't forget to add a "name" attribute to your labels as well. It should be the same as the field that the label is for.

```ruby
```

db/seeds.rb
```diff
@@ -5,3 +5,9 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+@hotel_1 = Hotel.create!(name: 'Stanly Hotel', location: 'Estes Park')
+@room_401 = @hotel_1.rooms.create!(rate: 50, suite: 'Room 401')
+@room_402 = @hotel_1.rooms.create!(rate: 50, suite: 'Room 402')
+@guest_1 = Guest.create!(name: 'Jeff', nights: 1)
+GuestRoom.create(guest: @guest_1, room: @room_401)
```
abdulredd, line 13: I don't think we really need seed data when you're doing an assessment since you'll spend most of your time in the tests and if you wanted to see what the page looks like you can run a `save_and_open_page`.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        expect(page).to have_content("Rate: 50")
+        expect(page).to have_content("Hotel: Stanly Hotel")
+      end
+
+      it 'Then I see a form to add a room to this guest' do 
+        visit "/guests/#{@guest_1.id}"
+
+        expect(page).to have_field(:room_id)
+        expect(page).to have_content("Add Room")
+      end
```
abdulredd, line 34: Just wanted to speak on these 2 lines here:

```ruby
expect(page).to have_content("Add Room")
```

After speaking with Chris, he said it's not necessary to have these in the tests moving forward since when we try to `fill_in` the field if it isn't on the page it would error out so just wanted to call that out here.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          fill_in :room_id, with: @room_402.id
+          click_on "Add Room"
+          expect(current_path).to eq("/guests/#{@guest_1.id}")
+          expect(page).to have_content("Rooms: Room 402")
+        end
+      end
+    end
+  end
+end
```
abdulredd, line 48: I think this is an excellent test, and fully covers what was in the user stories. Great job James.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      end 
+
+      it 'And I see a list of all the rooms they have stayed in as well as room fields and hotel' do
+        visit "/guests/#{@guest_1.id}"
+
+        expect(page).to have_content("Rooms stayed in")
+        expect(page).to have_content("Rooms: Room 401")
+        expect(page).to have_content("Rate: 50")
+        expect(page).to have_content("Hotel: Stanly Hotel")
+      end
```
abdulredd, line 27: The only thing I would add here is some `to_not` assertions as well to make sure the other room doesn't show up. This makes the test a little more robust.

```ruby
expect(page).to_not have_content("Rooms: Room 402")
```

You can also consider adding some `within` blocks as well:

```ruby
within "#rooms" do
  expect(page).to have_content("Rooms stayed in")
  expect(page).to have_content("Rooms: Room 401")
  expect(page).to have_content("Rate: 50")
  expect(page).to have_content("Hotel: Stanly Hotel")
  expect(page).to_not have_content("Rooms: Room 402")
end
```

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  describe 'relationships' do
+    it { should belong_to :guest }
+    it { should belong_to :room }
+  end
+end
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    it { should have_many :guest_rooms }
+    it { should have_many(:rooms).through(:guest_rooms)}
+  end
+end
```

spec/models/room_spec.rb
```diff
@@ -3,5 +3,7 @@
 RSpec.describe Room, type: :model do
   describe 'relationships' do
     it {should belong_to :hotel}
+    it { should have_many :guest_rooms }
+    it { should have_many(:guests).through(:guest_rooms)}
   end
 end
```

abdulredd:
Overall, a solid showing here James. Sucks you started with a few hiccups and were only able to get through 2 user stories. Keep up the great work!

PR #27, opened 2023-02-21
Jade stewart IC

jadekstewart3:


app/controllers/application_controller.rb
```diff
@@ -1,2 +1,8 @@
 class ApplicationController < ActionController::Base
+   add_flash_types :info, :error, :warning
```
abdulredd, line 2: You seem to always pull something interesting out of your bag... `add_flash_types`... I'm writing that down. Just be sure with all the "cool" stuff you make use of, you're understanding the benefits behind each. 🙂

app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  def create 
+    guest_room = GuestRoom.new(guest_id: params[:id], room_id: params[:room_id])
+
+    redirect_to "/guests/#{guest_room.guest_id}"
+
+    if guest_room.save
+      flash[:info] = "Room Successfully Added!"
+    else  
+      flash[:error] = "Error: #{error_message(guest_room.errors)}"
+    end
```
abdulredd, line 12: Great job adding some flash messages for sad paths.

app/views/rooms/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<%@rooms.each do |room|%>
+<div id="room_info-<%=room.id%>">
+  <p>Suite: <%=room.suite%></p>
+  <p>Rate: <%=room.rate%></p>
+  <p>Hotel: <%=room.hotel.name%></p>
+  <p>Number of Guests: <%=room.guests.count_guests%></p>
```
abdulredd, line 6: With `room.guests.count_guests`, while making it a "class" method on the Guest model is a good idea. You could have also added an instance method to the Room model as well which would have given you the ability to do `room.count_guests` in your view, which personally just reads a little cleaner to me, and a lot easier to test as well.

config/routes.rb
```diff
@@ -1,3 +1,7 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  get '/guests/:id',               to: 'guests#show'
+  get '/rooms',                    to: 'rooms#index'
+
+  post '/guests/:id/guest_rooms', to: 'guest_rooms#create'
```
abdulredd, line 6: The controller and action are perfect. You want the route, however, to be representative of the association we are creating if we are abiding by REST principles.

---

```ruby
end
```

db/migrate/20230221162117_create_guest_rooms.rb
```diff
@@ -0,0 +1,10 @@
+class CreateGuestRooms < ActiveRecord::Migration[5.2]
+  def change
+    create_table :guest_rooms do |t|
+      t.references :guest, foreign_key: true
+      t.references :room, foreign_key: true
+
+      t.timestamps
+    end
+  end
+end
```
abdulredd, line 10: Looking at this migration, I can almost say for certain that you are using the `rails g model` command. Kudos for picking that up! 🙌

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+ it { should have_many :rooms }
```
abdulredd, line 4: There are 2 associations in your `Guest` model, only one of them is being tested here. We need to also test for the one-to-many association between a Guest and GuestRooms. I also don't see the `#through` relationship test for `:rooms`.

```ruby
it { should have_many :guest_rooms }
it { should have_many(:rooms).through(:guest_rooms) }
```

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  describe ".scope" do 
+    it '.count_guests' do
```
abdulredd, line 7:
```ruby
describe "Scopes" do
  describe "::count_guests" do
    it "should return the total number of guests" do
```

The `it` block should explain what the method should do. Use the `describe` block to name what method is being tested.

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+RSpec.describe Guest, type: :model do
+ it { should have_many :rooms }
+
+  describe ".scope" do 
+    it '.count_guests' do 
+      Guest.create!(name: "Hannah", nights: 3)
+      Guest.create!(name: "Phil", nights: 3)
+      Guest.create!(name: "Denise", nights: 3)
+
+      expect(Guest.count_guests).to eq(3)
```
abdulredd, line 12: Unfortunately, this test does not fully cover the expected functionality of what is described in the user story. US3 asks for `the number of guests that have stayed in that room`. Technically this test makes sure that the method `returns the total number of guests`, but it is missing the modifier `for this Room`. 

This test would be ok if it had a Room also, and then adds each Guest here to it one by one, with expect statements between each. However, as written it is not covering the functionality expressed in the user story. 

```ruby
comfy_palace = Hotel.create!(name: "Comfy Palace", location: "Buena Vista")

comfy_presidential = comfy_palace.rooms.create!(rate: 100, suite: "Presidential") 

comfy_presidential.guests.create!(name: "Hannah", nights: 3)
comfy_presidential.guests.create!(name: "Phil", nights: 3)
comfy_presidential.guests.create!(name: "Denise", nights: 3)

expect(comfy_presedential.guests.count_guests).to eq(3)

comfy_presidential.guests.create!(name: "Jade", nights: 3)

expect(comfy_presedential.guests.count_guests).to eq(4)
```

We want to also make sure that our logic supports the assumption that our code isn't hard-coded. If we think about the TDD process, it says "write the amount piece of code to get the test to pass", in your assertion, the smallest amount would have been to just hard-code `3`, so we need to make the test more robust by adding an additional guest to the room and seeing the value change.

spec/models/room_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe Room, type: :model do
   describe 'relationships' do
-    it {should belong_to :hotel}
+    it { should belong_to :hotel }
+    it {should have_many :guest_rooms}
   end
```
abdulredd, line 7: You are missing one association test here.

```ruby
it { should belong_to :hotel }
it { should have_many :guest_rooms }
it { should have_many(:guests).through(:guest_rooms) } # <=====
```

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  let!(:hannah) { Guest.create!(name: "Hannah", nights: 3) }
+
+  let!(:hannah_room_1) {GuestRoom.create!(guest_id: hannah.id, room_id: room_1.id) }
+  let!(:hannah_room_2) {GuestRoom.create!(guest_id: hannah.id, room_id: room_2.id) }
+
+  describe 'Guest show page' do 
+    before (:each) do 
+      visit "/guests/#{hannah.id}"
+    end
```
abdulredd, line 18: Great job adding the `visit` to your `before :each`.

I'm always weary about using `let` with Rails, but as long as you understand the memoization piece behind it then I guess it's ok.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      within "#room_info-#{room_2.id}" do 
+        expect(page).to have_content("Suite: #{room_2.suite}\nRate: #{room_2.rate}\nHotel: Comfy Palace")
+      end
+    end
+
+    it 'can add a room to the guest' do 
+      fill_in 'room_id', with: "#{room_3.id}" 
+      click_on 'Submit'
+
+      expect(page).to have_content("Suite: #{room_3.suite}\nRate: #{room_3.rate}\nHotel: Comfy Palace")
```
abdulredd, line 39: This test is great but doesn't fully cover the entire user story:

```
...
Then I am redirected back to the guest's show page ❌ 
...
```

Just want to add an assertion to make sure the browser was properly redirected.

```ruby
expect(current_path).to eq('/guests/#{hannah.id}')
```

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it 'displays the rooms that the guest has stayed in with the rooms attributes' do
+
+      within "#room_info-#{room_1.id}" do 
+        expect(page).to have_content("Suite: #{room_1.suite}\nRate: #{room_1.rate}\nHotel: Comfy Palace")
+      end
+
+      within "#room_info-#{room_2.id}" do 
+        expect(page).to have_content("Suite: #{room_2.suite}\nRate: #{room_2.rate}\nHotel: Comfy Palace")
+      end
+    end
```
abdulredd, line 33: What I would have also added to this `it` block was an assertions to check that `room_3` did not show up on the page as well.

```ruby
expect(page).to_not have_content("Suite: #{room_3.suite}\nRate: #{room_3.rate}\nHotel: Comfy Palace")
```

spec/models/guest_room_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails_helper'
+
+RSpec.describe GuestRoom, type: :model do
+  it { should belong_to :room }
+  it { should belong_to :guest }
+end
```
abdulredd, line 6: 👍

abdulredd (commented):
Overall, a great job here Jade. You continue to produce excellent work, keep it up!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<div id="room_info-<%=room.id%>">
+  <p>Suite: <%=room.suite%></p>
+  <p>Rate: <%=room.rate%></p>
+  <p>Hotel: <%=room.hotel.name%></p>
+<%end%>
+
+<h5>Add Room:</h5>
+<%= form_tag("/guests/#{@guest.id}/guest_rooms", local: true)%>
+<%= text_field_tag(:room_id, params[:room_id])%>
+<%= submit_tag('Submit')%>
```
abdulredd, line 13: Great job setting up your form. Again, route should point to `"/guests/#{@guest.id}/rooms"`, but everything else is great. I would also add a `label_tag` as well.

PR #28, opened 2023-02-21
Antonio 2211

4D-Coder:
Completed Partially user story Two, time ended while fixing routes for the form.

This was incredibly difficult to complete after a long weekend and with morning brain. With having practiced every day, its so frustrating to be still having difficulty establishing proper workflow due to the sheer amount of nuances you have to know in order to build the features these stories request. That said, with some more time to practice this morning and then taking it after class today I feel like I would've gotten farther.

### DIFF
--- a/app/controllers/guest_rooms_controller.rb
+++ b/app/controllers/guest_rooms_controller.rb
@@ -1,0 +1,6 @@
+class GuestRoomsController < ActionController::Base
+  def create
+    GuestRoom.add_room(params)
+    redirect_to "guests/#{params["guest_id"]}"
+  end
+end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,0 +1,4 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+end
--- a/app/models/guest_room.rb
+++ b/app/models/guest_room.rb
@@ -1,4 +1,8 @@
 class GuestRoom < ApplicationRecord
   belongs_to :room
   belongs_to :guest
+
+  def self.add_room(params)
+    GuestRoom.create(room_id: params["room_id"], guest_id: params["guest_id"])
+  end
 end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,17 @@
+<section>
+  <h2><%= @guest.name %>'s Stay</h2>
+  <% @guest.rooms.each do |room| %>
+    <div class="<%"#{room.id}"%>">
+      <p>Hotel: <%= room.hotel.name %></p>
+      <p>Suite: <%= room.suite %></p>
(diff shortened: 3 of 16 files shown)


app/models/guest_room.rb
```diff
@@ -0,0 +1,8 @@
+class GuestRoom < ApplicationRecord
+  belongs_to :room
+  belongs_to :guest
+
+  def self.add_room(params)
+    GuestRoom.create(room_id: params["room_id"], guest_id: params["guest_id"])
```
cjsim89, line 6: This is fine but it's a bit extra/overkill, you can do `.create!` on methods in the corresponding controller. The only ones you are "required" to put in a model method would be anything outside of `.find`, `.find_by`, `.create, .update, .destroy, .all`. 

Also, note that if you make a model method, you'll need to test that model method. Since this method is really just doing a `.create`, and that comes packaged with Rails, we don't need to test it in the end.

app/controllers/guest_rooms_controller.rb
```diff
@@ -0,0 +1,6 @@
+class GuestRoomsController < ActionController::Base
+  def create
+    GuestRoom.add_room(params)
```
cjsim89, line 3: The setup here is great, I have a note on this method that it's a bit overkill to make this a model method, but it's fine this way too.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <h2><%= @guest.name %>'s Stay</h2>
+  <% @guest.rooms.each do |room| %>
+    <div class="<%"#{room.id}"%>">
+      <p>Hotel: <%= room.hotel.name %></p>
+      <p>Suite: <%= room.suite %></p>
+      <p>Nightly Rate: $<%= room.rate %></p>
+    </div>
+  <% end %>
+  <div class="add_room">
+    <%= form_with url: "/guests/#{@guest.id}/rooms", method: :post, local: true do |form| %>
```
cjsim89, line 11: Nice URL usage here! Very RESTful.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      it "Then I see a form to add a room to this guest" do
+        visit "/guests/#{@antonio.id}"
+        save_and_open_page
+
+          fill_in "room_id", with: "#{@executive.id}"
+          click_button "Submit"
+          
+        within("#{@executive.id}") do
```
cjsim89, line 43: Your tests overall look well-organized and strong; try to use a `to_not` inside a `within` block to be even stronger.

cjsim89 (commented):
Great work so far, just looking for more completion overall - all the practices you're showing in this exercise point to success, in your individual practice try to focus on speed & completion of all 3 user stories. The checklist/"cheat sheet" we talked about should help with this!

PR #29, opened 2023-02-21
Elle Majors

Elle-M:
Ran into several issues early on, but put down as much as I could. Issues: model relationships, guest_id in testing. Also, I realized I had misread the directions and created several hotels. By the time I realized this was when I looking at user story 3.

### DIFF
--- a/app/controllers/guest_controller.rb
+++ b/app/controllers/guest_controller.rb
@@ -1,0 +1,17 @@
+class GuestsController < ApplicationRecord
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  def update
+    room = Room.find(strong_params[:room_id])
+    room.update(guest_id: params[:id])
+
+    redirect_to "/guest/#{guest.id}"
+  end
+
+  private
+  def strong_params
+    params.require(:guest).permit(:room_id)
+  end
+end
--- a/app/controllers/room_controller.rb
+++ b/app/controllers/room_controller.rb
@@ -1,0 +1,9 @@
+class RoomsController < ApplicationRecord
+  def index
+    @rooms = Room.all
+  end
+
+  def show
+    @room = Room.find(params[:id])
+  end
+end
--- a/app/views/guests/show.html.erb
+++ b/app/views/guests/show.html.erb
@@ -1,0 +1,19 @@
+<h1>Guests <%= @guest.name %></h1>
+<hr>
+<div id="rooms">
+  <b>Guest <%= @guest.name %>'s Rooms:</b><br>
+  <% @guest.rooms.each do |room| %>
+  <ul> 
+  <li>Suite:<%= room.name %><li><br>
+  <li>Rate:<%= room.rate %><li><br>
+  <%# <li>Name:<%= room.hotel.name %><li><br> %>
+  <% end %>
+  </ul>
+
+  <div id="add-room">
+  <%= form_with model: @guest, local: true do |form| %>
+  <%= form.label :room_id %>
+  <%= form.text_field :room_id %>
+  <%= form.submit "Submit" %>
+  <% end %>
+</div>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :guests, only:[:show, :update]
+  resources :rooms, only: [:show]
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,3 +5,21 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+@hotel_1 = Hotel.create!(name: 'Echo Mountain Inn', city: 'Echo Mountain')
+@hotel_2 = Hotel.create!(name: 'Four Seasons', city: 'Vail')
+@hotel_3 = Hotel.create!(name: 'Ritz', city: 'Denver')
+
+@guest_1 = Guest.create!(name: 'Charlize Theron', nights: 3)
+@guest_2 = Guest.create!(name: 'Elle Majors', nights: 5 )
+@guest_3 = Guest.create!(name: 'Avery Bechtel', nights: 7)
+@guest_4 = Guest.create!(name: 'Tim Cook', nights: 2)
+
+@room_1 = @guest_1.room.create!(rate: 125, suite: 'Presidential', hotel_id: @hotel_1.id)
+@room_2 = @guest_2.room.create!(rate: 100, suite: 'Premium', hotel_id: @hotel_1.id)
+@room_3 = @guest_3.room.create!(rate: 300, suite: 'Executive', hotel_id: @hotel_2.id)
+@room_4 = @guest_4.room.create!(rate: 600, suite: 'VIP', hotel_id: @hotel_3.id)
+
+@hotel_rooms1 = HotelRoom.create!(hotel: @hotel_1 , room: @room_1)
+@hotel_rooms2 = HotelRoom.create!(hotel: @hotel_2, room: @room_2)
+@hotel_rooms3 = HotelRoom.create!(hotel: @hotel_3, room: @room_3)
--- a/spec/features/guests/show_spec.rb
+++ b/spec/features/guests/show_spec.rb
@@ -1,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.describe 'The Guest Show Page', type: :feature do
+  before do 
+    @hotel_1 = Hotel.create!(name: 'Echo Mountain Inn', location: 'Echo Mountain')
+    @hotel_2 = Hotel.create!(name: 'Four Seasons', location: 'Vail')
+    @hotel_3 = Hotel.create!(name: 'Ritz', location: 'Denver')
+  
+    @guest_1 = Guest.create!(name: 'Charlize Theron', nights: 3)
+    @guest_2 = Guest.create!(name: 'Elle Majors', nights: 5 )
+    @guest_3 = Guest.create!(name: 'Avery Bechtel', nights: 7)
+    @guest_4 = Guest.create!(name: 'Tim Cook', nights: 2)
+  
+    @room_1 = @guest_1.rooms.create!(rate: 125, suite: 'Presidential', hotel_id: @hotel_1.id)
+    @room_2 = @guest_2.rooms.create!(rate: 100, suite: 'Premium', hotel_id: @hotel_1.id)
+    @room_3 = @guest_3.rooms.create!(rate: 300, suite: 'Executive', hotel_id: @hotel_2.id)
+    @room_4 = @guest_4.rooms.create!(rate: 600, suite: 'VIP', hotel_id: @hotel_3.id)
+  
+    @hotel_rooms1 = HotelRoom.create!(hotel: @hotel_1 , room: @room_1)
+    @hotel_rooms2 = HotelRoom.create!(hotel: @hotel_2, room: @room_2)
+    @hotel_rooms3 = HotelRoom.create!(hotel: @hotel_3, room: @room_3)
+  end
+
+  describe 'user story 1' do
+    it 'shows the guests name' do
+      visit "/guest/#{guest_1.id}"  
+
+      expect(page).to have_content("Charlize Theron")
+      expect(page).to_not have_content("Tim Cook")
+    end
+
+    it 'displays a list of all the rooms they stayed in' do
+      visit "/guest/#{guest_1.id}"  
+
+      expect(page).to have_content("Presidential")
+    end
+
+    it 'includes the rooms suite, nightly rate, and the name of the hotel' do
+      visit "/guest/#{guest_1.id}"  
+
+      expect(page).to have_content("Suite: Presidential")
+      expect(page).to have_content("Rate: 125")
+      expect(page).to have_content("Name: Echo Mountain Inn")
+    end
+  end
+
+  describe 'user story 2' do
+    it 'displays a form to add a room to this guest' do
+      visit "/guest/#{guest_1.id}"  
+
+      within("#add-room") do
+        expect(page).to have_field("Add Room")
+        fill_in "Add Room", with: room_2.id 
+        click_button "Submit" 
+        # save_and_open
+      expect(current_path).to eq("/guest/#{guest_1.id}")
+      expect(page).to have_content("#{room_2.id}")
+      expect(page).to have_content("Premium")
+      end
+    end
+
+    it 'displays a form to add a room to this guest' do
+      visit "/guest/#{guest_1.id}"  
+
+      within("#add-room") do
+        expect(page).to have_field("Add Room")
+        fill_in "Add Room", with: room_5.id 
+        click_button "Submit" 
+      
+      expect(current_path).to eq("/guest/#{guest_1.id}")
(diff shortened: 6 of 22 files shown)


abdulredd:
So instead of making individual comments in all of your models I'll just go ahead and add all of the model comments here.

So what I want to start off with is the actual `schema.rb` file and take a look at what we were given and then move on to the `README` and take a look at what we had to do or what modifications we had to make to the database.

In the schema, we were given a one-to-many relationship between hotels and rooms where the rooms table contains a foreign key column for `hotel_id`. The model associations were already set up for us so we didn't have to do any additional work there for those two tables.

If we take a look at the model setup instructions from the `README` will see that we have four bullet points the first one saying that _"Hotels have many Rooms"_, the second one saying _"Rooms belong to a Hotel"_, again the database and associations have already been set up for us for these two tables. The only ones that we were left with were _"Guests have many Rooms"_ and then _"Rooms have many Guests"_.

Whenever you see a _"have many"_ to _"have many"_ (ie. many-to-many) type of relationship like this, this is a great indicator that you will be using a join table. You could call your join table GuestRooms or RoomGuests it doesn't really matter.

I saw that you created the Guests table which is great because we need the Guests table in order to join Guests to Rooms.

I also noticed that you did create a join table, but created a join table between Hotels and Rooms, again a join table is for when we have a many-to-many type of relationship. In this case, Hotels and Rooms have a one-to-many relationship, so no join table is needed there.

So now that we understand that we have to make a new join table, we need to make a new migration because we want to update the database. The command for that migration will look something like this:

`rails g migration create_guest_rooms guest:references room:references`

You'll see that I added two reference columns in the migration so that this join table will hold the foreign keys for those two tables.

Once you run the migration, migrate the database, and then create the `GuestRoom` model and then add the appropriate associations between a `Guest` and `GuestRoom`, a `Room` and `GuestRoom`, and finally a `Guest` to `Room` through `GuestRoom`.

I know that was a lot but I just wanted to be as thorough as possible so that way you won't be stuck with the setup for the next one. 🙂

spec/models/guest_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe Guest, type: :model do
+  describe 'relationships' do
+    # it {should have_many :hotels}
```
abdulredd, line 5: Because the requirements ask us to allow Guests to have many Rooms and for Rooms to have many Guests, we will need to have two associations here. One for having many GuestRooms and the other one for having many Rooms through GuestRooms.

```ruby
it { should have_many :guest_rooms }
it { should have_many(:rooms).through(:guest_rooms) }
```

spec/models/hotel_room_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe HotelRoom, type: :model do
```
abdulredd, line 3: We want this to be `GuestRoom` for the many-to-many relationship between Guests and Rooms.

The associations would be:

```ruby
it { should belong_to :guest }
it { should belong_to :room }
```

spec/models/hotel_spec.rb
```diff
@@ -3,5 +3,6 @@
 RSpec.describe Hotel, type: :model do
   describe 'relationships' do
     it {should have_many :rooms}
+    # it {should have_many(:guests).through :rooms }
```
abdulredd, line 6: No `:through` association required, since this is a one-to-many relationship.

spec/models/room_spec.rb
```diff
@@ -3,5 +3,8 @@
 RSpec.describe Room, type: :model do
   describe 'relationships' do
     it {should belong_to :hotel}
```
abdulredd, line 5: Because the requirements ask us to allow Guests to have many Rooms and for Rooms to have many Guests, we will need to have two associations here. One for having many GuestRooms and the other one for having many Guests through GuestRooms.

```ruby
it { should have_many :guest_rooms }
it { should have_many(:guests).through(:guest_rooms) }
```

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    @hotel_1 = Hotel.create!(name: 'Echo Mountain Inn', location: 'Echo Mountain')
+    @hotel_2 = Hotel.create!(name: 'Four Seasons', location: 'Vail')
+    @hotel_3 = Hotel.create!(name: 'Ritz', location: 'Denver')
+  
+    @guest_1 = Guest.create!(name: 'Charlize Theron', nights: 3)
+    @guest_2 = Guest.create!(name: 'Elle Majors', nights: 5 )
+    @guest_3 = Guest.create!(name: 'Avery Bechtel', nights: 7)
+    @guest_4 = Guest.create!(name: 'Tim Cook', nights: 2)
+  
+    @room_1 = @guest_1.rooms.create!(rate: 125, suite: 'Presidential', hotel_id: @hotel_1.id)
```
abdulredd, line 14: The error here: `ActiveModel::UnknownAttributeError: unknown attribute 'guest_id' for Room.`

This is because since there is no "in-between" join table, Rails assumes that there is a foreign key column called `guest_id` on the Rooms table, which there isn't one, so it is unable to associate the Guest and the Room you are attempting to create. 

If there was a GuestRooms join table, Rails would be able to associate the Guest and the Room you are creating here on the join table.

db/seeds.rb
```diff
@@ -5,3 +5,21 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+@hotel_1 = Hotel.create!(name: 'Echo Mountain Inn', city: 'Echo Mountain')
+@hotel_2 = Hotel.create!(name: 'Four Seasons', city: 'Vail')
+@hotel_3 = Hotel.create!(name: 'Ritz', city: 'Denver')
```
abdulredd, line 11: I think what you wanted instead of `city` was `location`. It happens to me all the time and I forget what the columns are called. I would recommend doing a split-view and keeping the schema file open so you can be sure you are using the correct columns.

db/seeds.rb
```diff
@@ -5,3 +5,21 @@
...
+
+@guest_1 = Guest.create!(name: 'Charlize Theron', nights: 3)
+@guest_2 = Guest.create!(name: 'Elle Majors', nights: 5 )
+@guest_3 = Guest.create!(name: 'Avery Bechtel', nights: 7)
+@guest_4 = Guest.create!(name: 'Tim Cook', nights: 2)
+
+@room_1 = @guest_1.room.create!(rate: 125, suite: 'Presidential', hotel_id: @hotel_1.id)
+@room_2 = @guest_2.room.create!(rate: 100, suite: 'Premium', hotel_id: @hotel_1.id)
+@room_3 = @guest_3.room.create!(rate: 300, suite: 'Executive', hotel_id: @hotel_2.id)
+@room_4 = @guest_4.room.create!(rate: 600, suite: 'VIP', hotel_id: @hotel_3.id)
```
abdulredd, line 21: `@guest_1.rooms.create!`... don't forget the `s`. The `rooms` that we are using here is the `has_many :rooms`  association that's on the `Guest` model.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    @room_4 = @guest_4.rooms.create!(rate: 600, suite: 'VIP', hotel_id: @hotel_3.id)
+  
+    @hotel_rooms1 = HotelRoom.create!(hotel: @hotel_1 , room: @room_1)
+    @hotel_rooms2 = HotelRoom.create!(hotel: @hotel_2, room: @room_2)
+    @hotel_rooms3 = HotelRoom.create!(hotel: @hotel_3, room: @room_3)
+  end
+
+  describe 'user story 1' do
+    it 'shows the guests name' do
+      visit "/guest/#{guest_1.id}"
```
abdulredd, line 26: Remember that since we are doing our setup in the `before :each`, we have to use an `@` sign in order to gain access to the guest:

`visit "/guests/#{@guest_1.id}" `

Also remember that, in REST the resource in the route is always plural. `/guests/`

It might even be better to put the `visit` in the `before :each` as well since you end up visiting the page for every test.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    @hotel_rooms2 = HotelRoom.create!(hotel: @hotel_2, room: @room_2)
+    @hotel_rooms3 = HotelRoom.create!(hotel: @hotel_3, room: @room_3)
+  end
+
+  describe 'user story 1' do
+    it 'shows the guests name' do
+      visit "/guest/#{guest_1.id}"  
+
+      expect(page).to have_content("Charlize Theron")
+      expect(page).to_not have_content("Tim Cook")
```
abdulredd, line 29: Great job adding a `expect(page).to_not have_content` as well in your test to make it more robust.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      expect(page).to have_content("Presidential")
+    end
+
+    it 'includes the rooms suite, nightly rate, and the name of the hotel' do
+      visit "/guest/#{guest_1.id}"  
+
+      expect(page).to have_content("Suite: Presidential")
+      expect(page).to have_content("Rate: 125")
+      expect(page).to have_content("Name: Echo Mountain Inn")
```
abdulredd, line 43: I'm a really big fan of these tests that not only test that the content shows up on the page but that there is some sort of like key that says this number `125` is associated with the `Rate:` and isn't just some random `125` that shows up anywhere on the page

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+  end
+
+  describe 'user story 2' do
+    it 'displays a form to add a room to this guest' do
+      visit "/guest/#{guest_1.id}"  
+
+      within("#add-room") do
```
abdulredd, line 51: Great job using the `within` block here!

Also, great job for adding an expectation for the page to have a field, but I think the field that we want to be checking for is `:room_id`. The only other thing I would add here is a check to make sure that the page also has a button because what is a form without a button or a way to submit that form?

`expect(page).to have_button("Submit")`

The last thing here is for the final 3 expectations, they don't need to be in the `within` block.

spec/features/guests/show_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        fill_in "Add Room", with: room_2.id 
+        click_button "Submit" 
+        # save_and_open
+      expect(current_path).to eq("/guest/#{guest_1.id}")
+      expect(page).to have_content("#{room_2.id}")
+      expect(page).to have_content("Premium")
+      end
+    end
+
+    it 'displays a form to add a room to this guest' do
```
abdulredd, line 62: This test looks like it's a duplicate.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    @hotel_rooms2 = HotelRoom.create!(hotel: @hotel_2, room: @room_2)
+    @hotel_rooms3 = HotelRoom.create!(hotel: @hotel_3, room: @room_3)
+  end
+
+  describe 'user story 3' do
+    it 'displays a list of all rooms' do
+      visit "/rooms"  
+
+      expect(page).to have_content("Presidential")
+      expect(page).to_not have_content("Premium")
```
abdulredd, line 29: `expect(page).to_not have_content("Premium")`

This portion of your test wouldn't actually pass because this Rooms index page will show all rooms in the database including the "Premium" room.

spec/features/rooms/index_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    it 'includes the rooms suite, nightly rate, and the name of the hotel' do
+      visit "/rooms"  
+
+      expect(page).to have_content("Suite: Presidential")
+      expect(page).to have_content("Rate: 125")
+      expect(page).to have_content("Suite:Premium")
+      expect(page).to have_content("Rate: 100")
+      expect(page).to have_content("Name: Echo Mountain Inn")
+    end
```
abdulredd, line 40: Great test!

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,19 @@
+<h1>Guests <%= @guest.name %></h1>
+<hr>
+<div id="rooms">
+  <b>Guest <%= @guest.name %>'s Rooms:</b><br>
+  <% @guest.rooms.each do |room| %>
+  <ul> 
+  <li>Suite:<%= room.name %><li><br>
+  <li>Rate:<%= room.rate %><li><br>
+  <%# <li>Name:<%= room.hotel.name %><li><br> %>
```
abdulredd, line 9: This line should be all you need to get the name of the hotel.

config/routes.rb
```diff
@@ -1,3 +1,5 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :guests, only:[:show, :update]
+  resources :rooms, only: [:show]
```
abdulredd, line 4: Great job attempting to implement the `resources` syntax, there are only a few that aren't implemented according to the user stories.

User story 2 is for the Rooms index page, in your `routes.rb` you have a show route.

For user story 3, this is an association between a room and a guest, so you wither want a nested association route:

```ruby
end

resources :rooms, only: :index
```

or 

```ruby
resources :guests, only: :show
resources :rooms, only: index
resources :guest_rooms, only: :create
```

The biggest thing I want to emphasize is that we are adding using a `POST` to `create` because we want to create an association between a Guest and a Room.

app/views/guests/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <% end %>
+  </ul>
+
+  <div id="add-room">
+  <%= form_with model: @guest, local: true do |form| %>
```
abdulredd, line 14: This wouldn't point to the `@guest` model because we want to create an association between a Guest and a Room.

```ruby
<%= form_with url: "/guests/#{@guest.id}/rooms", local: true do |form| %>
```

or 

```ruby
<%= form_with url: "/guests_rooms", local: true do |form| %>
    <%= form.hidden :guest_id, value: @guest.id %>
```

app/controllers/guest_controller.rb
```diff
@@ -0,0 +1,17 @@
+class GuestsController < ApplicationRecord
+  def show
+    @guest = Guest.find(params[:id])
+  end
+
+  def update
```
abdulredd, line 6: We aren't exactly updating the guest, but instead creating a new record in the GuestRooms table. You wouldn't need this action in this controller.

The action, `create` would be defined in your `GuestRoomsController` with something like this:

```ruby
def create
  guest = Guest.find(params[:id]) # or guest = Guest.find(params[:guest_id]), depending on how you set up your routes
  room = Room.find(params[:room_id])

  GuestRoom.create!(guest_id: guest.id, room_id: room.id)

  redirect_to "/guests/#{guest.id}"
end
```

app/controllers/room_controller.rb
```diff
@@ -0,0 +1,9 @@
+class RoomsController < ApplicationRecord
+  def index
+    @rooms = Room.all
+  end
+
+  def show
```
abdulredd, line 6: Because there isn't a user story that deals with showing a single room, we wouldn't need this `show` action.

abdulredd:
One final thought. You do not need to worry about creating seed data when working on these IC's since we are dealing with the tests for the most part and therefore the test database. You can, if you want, add seed database if you plan on firing up your application with `rails s`. Other than that, you can use a `save_and_open_page` in your tests if you want to see how the page looks in your browser.

